PR #36, opened 2019-07-12
Ops review

amycogniac:


cogniac/ops_review.py
```diff
@@ -0,0 +1,173 @@
...
+        media_id(String):                Media_id associated with the result if inspection is done on per media basis
+        domain_unit(String):             Domain_unit associated with the result if inspection is done on per part basis
+        """
+        data = {}
+        if review_unit:
+            data['review_unit'] = review_unit
+        if review_items:
+            data['review_items'] = review_items
+
+        resp = connection._post("/opsReview", json=data)
```
deepshikha-jha: I think the url here is /ops/review instead of /opsReview

cogniac/cogniac.py
```diff
@@ -158,10 +158,10 @@ def __init__(self,
         self.tenant = CogniacTenant.get(self)
         self.user = CogniacUser.get(self)
 
-        if self.tenant.region is not None:
-            # use tenant object's specified region preference
-            print "Using API endpoint from Tenant:", self.tenant.region
-            self.url_prefix = 'https://' + self.tenant.region + '/1'
+        #if self.tenant.region is not None:
```
wskish: Do you intend for this to stay commented out?

deepshikha-jha:
The scheduler.py has a function called create_ops_review. You currently don't have this function on the cogniac.py

amycogniac:
No, I was using staging for the branch. We cannot use staging otherwise.
Should we allow people to set to use staging for bnsf?

amycogniac:
Merge final version, ready to commit

PR #43, opened 2019-11-24
bare minimum to allow trigger

amycogniac:


wskish:
implemented elsewhere

PR #44, opened 2019-11-27
Netcam support

amycogniac:


cogniac/cogniac.py
```diff
@@ -416,6 +418,20 @@ def get_version(self, auth=False):
         raise_errors(resp)
         return resp.json()
 
+    def get_all_cameras(self):
+        """
+        return CogniacApplications for all applications belonging to the currently authenticated tenant
```
sergii-bond: the function description is not right

cogniac/cogniac.py
```diff
@@ -416,6 +418,20 @@ def get_version(self, auth=False):
...
+        """
+        return CogniacNetworkCamera.get_all(self)
+
+    def get_camera(self, network_camera_id):
+        """
+        return an existing CogniacApplication
```
sergii-bond: the function description is not right

PR #61, opened 2020-04-06
Add gateway support in CloudCore and local APIs.

mgub:
### Changes
- Refactored `CogniacGateway` object to include APIs to send events to EF via CloudCore.
- Added `get_all_gateways` and `get_gateways` functions to `CogniacConnection` object.
- Added optional `url_prefix` parameter to `get_gateway` to use a local gateway. If no `url_prefix` is specified, local EF API calls will fail.
- Allow pure local interaction with EF by directly instantiating `CogniacGateway` with `url_prefix`. In this case, CloudCore APIs will fail.

### Next
- Don't add APIs specific to `network_interfaces` in this PR, create a follow-up issue to discuss. Suggest creating additional, distinct class for network interfaces such as `NetworkInterface`.
- Testing for different API usage cases introduced here.
- API Documentation and examples.

### DIFF
--- a/cogniac/cogniac.py
+++ b/cogniac/cogniac.py
@@ -21,7 +21,7 @@ from subject import CogniacSubject
 from tenant  import CogniacTenant
 from user    import CogniacUser
 from media   import CogniacMedia
-from gateway import CogniacGateway
+from edgeflow import CogniacEdgeFlow
 
 from network_camera import CogniacNetworkCamera
 
@@ -439,21 +439,21 @@ class CogniacConnection(object):
         """
         return CogniacNetworkCamera.get(self, network_camera_id)
 
-    def get_all_gateways(self):
+    def get_all_edgeflows(self):
         """
-        return CogniacGateway for all netcams belonging to the currently authenticated tenant
+        return CogniacEdgeFlow for all netcams belonging to the currently authenticated tenant
         """
-        return CogniacGateway.get_all(self)
+        return CogniacEdgeFlow.get_all(self)
 
-    def get_gateway(self, gateway_id, url_prefix=None):
+    def get_edgeflow(self, edgeflow_id, url_prefix=None):
         """
-        return an existing CogniacGateway 
+        return an existing CogniacEdgeFlow 
 
-        gateway_id (String):  The id of the Cogniac Gateway to return
-        url_prefix (String):  The local URL prefix of a Gateway
+        edgeflow_id (String):  The id of the Cogniac EdgeFlow to return
+        url_prefix (String):  The local URL prefix of a EdgeFlow 
         """
-        return CogniacGateway.get(connection=self,
-                                  gateway_id=gateway_id,
+        return CogniacEdgeFlow.get(connection=self,
+                                  edgeflow_id=edgeflow_id,
                                   url_prefix=url_prefix)
 
 
--- a/cogniac/cogniac.py
+++ b/cogniac/cogniac.py
@@ -21,6 +21,7 @@ from subject import CogniacSubject
 from tenant  import CogniacTenant
 from user    import CogniacUser
 from media   import CogniacMedia
+from gateway import CogniacGateway
 
 from network_camera import CogniacNetworkCamera
 
@@ -437,6 +438,23 @@ class CogniacConnection(object):
         network_camera_id (String):  The id of the Cogniac network camera to return
         """
         return CogniacNetworkCamera.get(self, network_camera_id)
+
+    def get_all_gateways(self):
+        """
+        return CogniacGateway for all netcams belonging to the currently authenticated tenant
+        """
+        return CogniacGateway.get_all(self)
+
+    def get_gateway(self, gateway_id=None, url_prefix=None):
+        """
+        return an existing CogniacGateway 
+
+        gateway_id (String):  The id of the Cogniac Gateway to return
+        url_prefix (String):  The local URL prefix of a Gateway
+        """
+        return CogniacGateway.get(self,
+                                  gateway_id=gateway_id,
+                                  url_prefix=url_prefix)
 
 
 if __name__ == "__main__":
--- a/cogniac/cogniac.py
+++ b/cogniac/cogniac.py
@@ -441,20 +441,17 @@ class CogniacConnection(object):
 
     def get_all_edgeflows(self):
         """
-        return CogniacEdgeFlow for all netcams belonging to the currently authenticated tenant
+        return CogniacEdgeFlow for all EdgeFlows belonging to the currently authenticated tenant
         """
         return CogniacEdgeFlow.get_all(self)
 
-    def get_edgeflow(self, edgeflow_id, url_prefix=None):
+    def get_edgeflow(self, edgeflow_id):
         """
         return an existing CogniacEdgeFlow 
 
         edgeflow_id (String):  The id of the Cogniac EdgeFlow to return
-        url_prefix (String):  The local URL prefix of a EdgeFlow 
         """
-        return CogniacEdgeFlow.get(connection=self,
-                                  edgeflow_id=edgeflow_id,
-                                  url_prefix=url_prefix)
+        return CogniacEdgeFlow.get(self, edgeflow_id)
 
 
 if __name__ == "__main__":
--- a/cogniac/cogniac.py
+++ b/cogniac/cogniac.py
@@ -445,14 +445,14 @@ class CogniacConnection(object):
         """
         return CogniacGateway.get_all(self)
 
-    def get_gateway(self, gateway_id=None, url_prefix=None):
+    def get_gateway(self, gateway_id, url_prefix=None):
         """
         return an existing CogniacGateway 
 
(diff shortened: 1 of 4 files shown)


cogniac/gateway.py
```diff
@@ -35,23 +50,88 @@ class CogniacGateway(object):
...
+        get single network camera
+        connnection (CogniacConnection): Authenticated CogniacConnection object
+        gateway_id (String): the unique identifier of a Gateway object
+        returns CogniacGateway object
+        """
+        gateway_dict = None
+        if connection and gateway_id:
+            resp = connection._get("/gateways/{}".format(gateway_id))
+            gateway = resp.json()
+        return CogniacGateway(connection=connection, url_prefix=url_prefix, gateway_dict=gateway)
```
spencer-kt, line 72: `gateway` undefined if conditional not meet.  Line 68 should be `gateway = None`, right?

cogniac/gateway.py
```diff
@@ -35,23 +50,88 @@ class CogniacGateway(object):
...
         if 'COG_GW_URL_PREFIX' in os.environ:
             url_prefix = os.environ['COG_GW_URL_PREFIX']
 
-        if not url_prefix:
-            raise Exception("No EdgeFlow URL prefix was specified.")
+        if not connection and not url_prefix:
+            raise Exception("A URL must be specified for either a CloudCore or EdgeFlow API.")
+
+        if connection and not gateway_dict:
+            raise Exception("Missing gateway object.")
```
spencer-kt, line 104: This logic doesn't account for `connection=None` cases.  Does this break if it's None?

mgub:
Thanks for taking a look, Spence. I'm still working on it and thinking
about a couple things... just a heads up. I'm hoping to wrap it up today
though.

One if the questions I'm thinking about is whether network interfaces
should have their own functions, or if accessing the object is sufficient
for now.

cogniac/cogniac.py
```diff
@@ -438,6 +439,23 @@ def get_camera(self, network_camera_id):
         """
         return CogniacNetworkCamera.get(self, network_camera_id)
 
+    def get_all_gateways(self):
```
spatel-cog, line 442: edgeflows instead of gateways?

cogniac/gateway.py
```diff
@@ -35,23 +50,88 @@ class CogniacGateway(object):
...
+    #  get
+    ##
+    @classmethod
+    @retry(stop_max_attempt_number=8, wait_exponential_multiplier=500, retry_on_exception=server_error)
+    def get(cls,
+            connection=None,
+            gateway_id=None,
+            url_prefix=None):
+        """
+        get single network camera
```
spatel-cog, line 63: get single gateway

cogniac/gateway.py
```diff
@@ -35,23 +50,88 @@ class CogniacGateway(object):
...
+        connnection (CogniacConnection): Authenticated CogniacConnection object
+        gateway_id (String): the unique identifier of a Gateway object
+        returns CogniacGateway object
+        """
+        gateway_dict = None
+        if connection and gateway_id:
+            resp = connection._get("/gateways/{}".format(gateway_id))
+            gateway = resp.json()
```
spatel-cog, line 71: is there an exception that should be handled from the connection._get call?

cogniac/gateway.py
```diff
@@ -35,23 +50,88 @@ class CogniacGateway(object):
...
+    def __init__(self, connection=None, gateway_dict=None, timeout=60, url_prefix=None):
         """
         Initialize a CogniacGateway object.
         
         url_prefix (String):          URL prefix for a Cogniac EdgeFlow device.
         """
+        if gateway_dict:
+            super(CogniacGateway, self).__setattr__('_gateway_keys', gateway_dict.keys())
+
         if 'COG_GW_URL_PREFIX' in os.environ:
```
spatel-cog, line 97: CORE_FQDN

cogniac/gateway.py
```diff
@@ -35,23 +50,88 @@ class CogniacGateway(object):
...
+    #  get_all
+    ##
+    @classmethod
+    def get_all(cls, connection):
+        """
+        return all CogniacGateway objects belonging to the currently authenticated tenant
+
+        connnection (CogniacConnection):     Authenticated CogniacConnection object
+        """
+        resp = connection._get('/tenants/%s/gateways' % connection.tenant.tenant_id)
```
spatel-cog, line 84: any exception handling required?

cogniac/gateway.py
mgub: I'm thinking that it should be possible to directly instantiate the `CogniacGateway` to connect to an EdgeFlow without first authenticating with a Core.

The idea here was that `gateway_id` can be `None` and `CogniacGateway` can be instantiated with only `url_prefix` for the EF's local API. This seems useful for cases in which a user wants to use the SDK to upload images to an EdgeFlow. Note, in this case, the Core APIs wouldn't be available, so first creating a `CogniacConnection` object seems unnecessary.

I'm trying to push validation logic into the `CogniacGateway` initialize to ensure that the `CogniacGateway` is valid for requests against (1) local EF API only, (2) CloudCore APIs only, (3) both.

Does this seem reasonable?

FWIW, I think the committed code handles these cases, but it may not be clear without some examples. I will add those to the PR after testing.

cogniac/gateway.py
mgub: See https://github.com/Cogniac/cogniac-sdk-py/pull/61#discussion_r404443512.

I'm thinking that `connection` can be `None` if using only the local APIs, so the SDK could be used on a LAN with the EF, potentially without access to a Core (e.g., CloudCore via WAN).

cogniac/cogniac.py
mgub: Yeah, that'd be better. I'll make those changes.

cogniac/gateway.py
mgub: I thought this followed the existing pattern, but I'll verify.

cogniac/gateway.py
mgub: Should I assume API version 1? Is there an environment variable for that or can we choose one, if not?

mgub: `COG_GW_URL_PREFIX` actually refers to the EF's HTTP API, to enable API access to `/version`, `/process`, etc., so it's distinct from `CORE_FQDN` and `COG_URL_PREFIX`. There's probably a better name for it, though.

cogniac/gateway.py
mgub: Other entities' `get_all_` functions also follow this pattern.

cogniac/gateway.py
wskish: to the extent this needs to be configured it should come from CloudCore not the environment.  Lets hardcode it to 1 for now and if it changes we will figure out how to communicate that via cloudcore config.

mgub: I removed `COG_GW_URL_PREFIX` here. It refers to the address of the EdgeFlow's API, not CloudCore, which I believe isn't currently available via CloudCore.

mgub:
Here are some usage examples to get a feel for the current state of this branch:

```
# Get all tenant EdgeFlows.
cc = CogniacConnection()
edgeflows = cc.get_all_edgeflows()
# pprint(edgeflows)

# Case 1: CloudCore APIs only.
cc = CogniacConnection()
edgeflow = cc.get_edgeflow(EDGEFLOW_ID)
pprint(edgeflow)
edgeflow.reboot()
exit()

# Case 2: EdgeFlow APIs only.
from cogniac import CogniacEdgeFlow
edgeflow = CogniacEdgeFlow(url_prefix="http://192.168.1.130:8000/1")
edgeflow_version = edgeflow.get_version()
print(edgeflow_version)
# edgeflow.reboot() # fails
exit()

# Case 3: CloudCore and EdgeFlow APIs.
cc = CogniacConnection()
edgeflow = cc.get_edgeflow(EDGEFLOW_ID, url_prefix="http://192.168.1.230:8000/1")
edgeflow_version = edgeflow.get_version()
# media_detections = edgeflow.process_media(subject_uid=SUBJECT_UID,
#                                          filename=IMAGE_FILEPATH)
print(edgeflow_version)
# edgeflow.flush_upload_queue()
# edgeflow.factory_reset()
# edgeflow.upgrade()
edgeflow.reboot()
# edgeflow.status()
exit()
```

mgub:
In the above examples, `CogniacEdgeFlow` is both the gateway object/client (CloudCore) and a client interface to a local EdgeFlow. I'm considering separating this into two classes, `EdgeFlow` for the object/client interface, and a separate `EdgeFlowClient` for local access.

I think this may clarify the purpose and prevent overloading the single class. Any preferences on that?

amycogniac:
how about delete gateway? can we add it to make it easier?

wskish:
Gateway is going to be deleted by itself as a side effect of factory reset. https://github.com/Cogniac/gateway/issues/166

cogniac/gateway.py
mgub: The IP address should be taken from the gateway object.

PR #63, opened 2020-05-08
Support for user token requests in CogniacConnection.

mgub:
In some cases, SDK clients try to make requests that require only user token and specifying a specific tenant token is undesirable. However, currently creating a `CogniacConnection` requires specifying a tenant.

This change allows a `CogniacConnection` to be created. If a `tenant_id` is present, the tenant token will be used in requests, the same as before. If no `tenant_id` is present, requests use the user token.


wskish:
What is the use case driving this?  I am concerned this is a significant deviation from current behavior.

PR #67, opened 2020-06-25
add support for request_time_region_upload

amycogniac:


cogniac/edgeflow.py
```diff
@@ -222,6 +222,17 @@ def upload():
...
 
+    def request_time_region_upload(self, start_time, stop_time):
+        """
+        Request a time region to upload iamges from the EdgeFlow to CloudCore
+        Must specify start_time and stop_time in unix time
+        """
+        args = dict()
+        args['start_time'] = start_time
+        args['stop_time'] = stop_time
+        self._cc._post("/gateways/%s/event/request_time_region_upload" % (self.gateway_id), data=args)
```
mgub, line 233:
```suggestion
        self._cc._post("/gateways/%s/event/request_time_region_upload" % (self.gateway_id), json=args)
```

mgub, line 233:
```suggestion
        self._cc._post("/gateways/%s/event/time_bound_media_upload" % (self.gateway_id), data=args)
```

amycogniac:
ok, now works. I require a start_time and end_time for upload request since otherwise user can easily upload everything

cogniac/edgeflow.py
```diff
@@ -222,16 +222,28 @@ def upload():
     #
     # -------------------------------------------------------------------------
 
-    def flush_upload_queue(self, start_time=None, stop_time=None):
+    def time_bound_media_upload(self, start_time, end_time):
```
wskish, line 225: It might be convenient to have an end_time=None default under which condition we set as current time() for user.

cogniac/edgeflow.py
```diff
@@ -222,16 +222,28 @@ def upload():
...
         Flush the EdgeFlow to Core upload queue
         """
         args = dict()
         if start_time is not None:
             args['start_time'] = start_time
-        if stop_time is not None:
-            args['stop_time'] = stop_time
-        self._cc._post("/gateways/%s/event/flush_upload_queue" % (self.gateway_id), data=args)
+        if end_time is not None:
+            args['end_time'] = end_time
```
mgub, line 244: Both `start_time` and `end_time` were updated to be required in the public API, so they should be set in the request.

If you can initialize default values in this function, they can be made optional in the SDK (e.g., set default value `end_time=time()` or initialize like `end_time = time() if end_time is not None else end_time`).

PR #71, opened 2020-09-03
add support for set_boot_software_version

amycogniac:
right now there is no status reporting for the set_boot_software_version from edgeflow. I created an issue for edgeflow_upgrade

### DIFF
--- a/cogniac/edgeflow.py
+++ b/cogniac/edgeflow.py
@@ -261,7 +261,7 @@ class CogniacEdgeFlow(object):
             args['software_version'] = software_version
         self._cc._post("/gateways/%s/event/upgrade" % (self.gateway_id), json=args)
 
-    def rollback(self, software_version):
+    def set_boot_software_version(self, software_version):
         """
         Upgrade the edgeflow to the specified software version
         """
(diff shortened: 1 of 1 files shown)


cogniac/edgeflow.py
```diff
@@ -259,7 +259,16 @@ def upgrade(self, software_version):
         args = dict()
         if software_version is not None:
             args['software_version'] = software_version
-        self._cc._post("/gateways/%s/event/upgrade" % (self.gateway_id), data=args)
+        self._cc._post("/gateways/%s/event/upgrade" % (self.gateway_id), json=args)
```
mgub, line 262: Just a comment—it's probably better to set `json=` than `data=` for API requests in the SDK unless JSON is decidedly not the format being exchanged for a given endpoint. Using it sets the headers correctly.

mgub (approved):

cogniac/edgeflow.py
```diff
@@ -259,7 +259,16 @@ def upgrade(self, software_version):
...
+
+    def rollback(self, software_version):
```
wskish, line 264: how about we mirror the endpoint name w/ set_boot_software_version()?   this can be used to roll back or forward as long as the software is downloaded.

wskish (changes requested):

PR #74, opened 2020-11-03
Add set_assignment option to media.

mgub:
Can `force_set` be removed?


wskish:
lets add comment to effect force_set is deprecated and open an issue to remove force_set with next major version bump

wskish (approved):

PR #80, opened 2020-12-03
upload genicam xml, fix #79

### LINKED ISSUE #79: Upload genicam xml
Create a function in CogniacNetworkCamera to upload genicam xml.

sergii-bond:
Here is a test case:
```
import cogniac
from cogniac.network_camera import CogniacNetworkCamera

cc = cogniac.CogniacConnection(tenant_id="your_tenant")

cam_name = "test_cam"

for c in cc.get_all_cameras():
    if c.camera_name == cam_name:
        cam = c
        print("Found a test camera")
        break
else:
    print("Creating a test camera")
    cam = CogniacNetworkCamera.create(cc, name=cam_name, url=u'gigv://169.254.23.111/')

test_xml = "test.xml"

with open(test_xml, "w") as fp:
    fp.write("some xml info")

cam.upload_genicam_xml(test_xml)
```

### DIFF
--- a/cogniac/network_camera.py
+++ b/cogniac/network_camera.py
@@ -6,6 +6,7 @@ Copyright (C) 2019 Cogniac Corporation
 
 from retrying import retry
 import sys
+import os
 from common import server_error
 
 camera_model_keys = ['last_pose_change_timestamp',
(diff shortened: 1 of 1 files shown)


sergii-bond:
And the test case returns `cogniac.common.ClientError: ClientError (403): {"message":"Permission Denied"}`. Is there an issue with api ?

wskish (approved):

mgub:
@sergii-bond Currently that endpoint requires the "ef_device" role, which is used by EF machine users created during EF setup.

What user/tenant roles are associated with the account you're using?

sergii-bond:
@mgub I was using a "Cogniac - Measurement" tenant_id=r2um06f1d0mr, and my local COG_USER credentials. So obviously not related to 'ef_device' role. I don't have a way to test then, could you please help to verify the code ? I also think I don't have access to s3, so won't be able to check if an xml file ends up there.

wskish:
thinking about this some more; this code probably isn't a good fit for the sdk.  Let's move it to the edgeflow auto discovery side.

wskish (changes requested):
move to autodiscovery

PR #82, opened 2020-12-03
app_type_config as object, fix #73

### LINKED ISSUE #73: change in app.app_type_config is not saved automatically
In cogniac application object, if there is a change in an attribute value, the attribute value should be saved automatically, as specified in line:
https://github.com/Cogniac/cogniac-sdk-py/blob/764f360cfeac301bc3583fab397b40a4f3586e50/cogniac/app.py#L145

However, app_type_config is a dictionary. Changing a key-value insides the app_type_config dict does not change the id (or pointer) to app_type_config dict. Hence, the above set_attr is not triggered.

To reproduce,
```python
app_id = "some app_id"
app = cc.get_application(app_id)
app.app_type_config['fps_requirement'] = 2
app = cc.get_application(app_id)
```  
You will see the change is not saved. However if you do the following, the change is saved:
```python
import copy
app_id = "some app_id"
app = cc.get_application(app_id)
app.app_type_config['fps_requirement'] = 2
app.app_type_config = copy.deepcopy(app.app_type_config)
app = cc.get_application(app_id)
```

sergii-bond:
@wskish please check if this implementation is according to your expectations, as you commented in #73.
Such operation as ```app.app_type_config.max_boxes = 500``` would post the result and change the app's configuration.

### DIFF
--- a/cogniac/app.py
+++ b/cogniac/app.py
@@ -5,7 +5,7 @@ Copyright (C) 2016 Cogniac Corporation
 """
 
 from retrying import retry
-from common import *
+from common import server_error
 
 
 class CogniacApplication(object):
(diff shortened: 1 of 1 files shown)


wskish (approved):

wskish:
looks good, thanks for taking care of this one

PR #90, opened 2021-04-01
__str__ must return type str

### LINKED ISSUE #89: Subjects __repr__ and __str__ methods cause TypeError
Found this because calling CogniacConnection.get_all_subjects() causes:

`TypeError: __repr__ returned non-string (type bytes)`

I believe it's these lines, the Application object doesn't have this problem, and returns the string directly:
https://github.com/Cogniac/cogniac-sdk-py/blob/11f985a07bbaf5616315abbb0a56c75bc7928472/cogniac/subject.py#L216
https://github.com/Cogniac/cogniac-sdk-py/blob/11f985a07bbaf5616315abbb0a56c75bc7928472/cogniac/subject.py#L220

Perhaps related to Python 2 to 3 compatibility?

sergii-bond:
resolve #89

Didn't find any other way than by checking python version.

### DIFF
--- a/cogniac/app.py
+++ b/cogniac/app.py
@@ -169,7 +169,8 @@ class CogniacApplication(object):
         super(CogniacApplication, self).__setattr__(name, value)
 
     def __repr__(self):
-        return "%s (%s)" % (self.name, self.application_id)
+        s = "%s (%s)" % (self.name, self.application_id)
+        return s if (sys.version_info.major > 2) else s.encode(sys.stdout.encoding)
 
     def add_output_subject(self, subject):
         """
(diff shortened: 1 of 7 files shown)


sergii-bond:
please check Sean's comment https://github.com/Cogniac/cogniac-sdk-py/issues/89#issuecomment-812089425, using 'six' library is another way. We also have dict iteritems() in app.py which in python3 is replaced by items(), 'six' library claims to address that as well.

wskish:
six does seem to popular for dealing with these cases; lets go with the six approach

sergii-bond:
I'll let Sean make a pull request with six, he is working on that.

PR #91, opened 2021-04-01
fix: Add six library for python_2_unicode_compatible class decorator

CogniacViSean:
This allows for py2 and py3 compatible __str__() methods

### DIFF
--- a/cogniac/app.py
+++ b/cogniac/app.py
@@ -5,9 +5,11 @@ Copyright (C) 2016 Cogniac Corporation
 """
 
 from retrying import retry
+import six
 from .common import server_error
 
 
+@six.python_2_unicode_compatible
 class CogniacApplication(object):
     """
     CogniacApplication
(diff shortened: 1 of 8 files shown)


cogniac/app.py
```diff
@@ -168,9 +168,12 @@ def __setattr__(self, name, value):
 
         super(CogniacApplication, self).__setattr__(name, value)
 
-    def __repr__(self):
+    def __str__(self):
```
sergii-bond, line 171: does this class need @six.python_2_unicode_compatible too ?

sergii-bond (changes requested):

cogniac/app.py
CogniacViSean: I did mean to include it, I don't think it strictly requires it because it originally only had a __repr__() without encoding the string for printing, but might as well include for potential unicode compatibility and consistency with the other classes?

sergii-bond: The app name can contain non-ascii characters, I think it should be similar to subject class.

sergii-bond (approved):

CogniacViSean:
Closes #89

PR #94, opened 2021-05-27
Add support setting "Sidelined" consensus type on a subject media association.

mgub:
Resolves https://github.com/Cogniac/CloudCore/issues/167.

This supports the addition of the "Sidelined" consensus type for applications to enable application reviewers to "sideline" media when providing feedback, meant to indicate the media is impossible to review or label with certainty (e.g., image is corrupt or somehow doesn't represent a typical image for an application).

### DIFF
--- a/cogniac/app.py
+++ b/cogniac/app.py
@@ -230,7 +230,7 @@ class CogniacApplication(object):
         subjects (list of dicts):      Subject-media association dictionaries of the form:
 
             subject_uid:               Subject UID
-            result (str):              Either 'True', 'False', 'Uncertain'
+            result (str):              One of 'True', 'False', 'Uncertain', 'Sidelined'
             app_data_type (String):    (Optional) Type of extra app-specific data for certain app types
             app_data (Object):         (Optional) Additional, app-specific, subject-media association data
 
--- a/cogniac/media.py
+++ b/cogniac/media.py
@@ -400,8 +400,8 @@ class CogniacMedia(object):
                         'False' if there is consensus that the subject is not associated w/the media
                             (Media will be used as a negative training example of the subject.)
                          None if if there is not enough evidence to reach consensus
-                         'Sidelined' if there is consensus that the media could not be considered
-                            associated with the subject.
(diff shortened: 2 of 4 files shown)


cogniac/media.py
```diff
@@ -394,12 +394,14 @@ def subjects(self):
...
         app_data_type	optional app data type if applicable
         app_data        optional app data if applicable
-        consensus		'True', 'False', or None
+        consensus		'True', 'False', 'Sidelined', or None
                         'True' if there is consensus that the subject is associated with the media
                             (Media will be used as a positive training example of the subject)
                         'False' if there is consensus that the subject is not associated w/the media
                             (Media will be used as a negative training example of the subject.)
                          None if if there is not enough evidence to reach consensus
+                         'Sidelined' if there is consensus that the media could not be considered
```
wskish, line 403: 'Sidelined' if there is consensus that the media should not be used for training applications dependent on the subject.

mgub: Thanks.

cogniac/app.py
```diff
@@ -230,7 +230,7 @@ def post_feedback(self, media_id, subjects):
         subjects (list of dicts):      Subject-media association dictionaries of the form:
 
             subject_uid:               Subject UID
-            result (str):              Either 'True', 'False', 'Uncertain'
+            result (str):              One of 'True', 'False', 'Uncertain', 'Sidelined'
```
wskish, line 233: do we sill support "uncertain" in the backend?  should probably remove references to here here

wskish (changes requested):

cogniac/app.py
mgub: My impression was that support for it was incomplete and it's not actually being used. I saw scattered references to it, but haven't come across business logic that consumes it for anything in particular. Seems like it may have been in the works as something like what Sidelining will do.

I can take a look. It would be nice to disallow that value and remove reference to it clarify the feedback pathway if it really isn't being used.

wskish (approved):

PR #96, opened 2021-05-28
handle file pointer based media uploads when size > 12MB

spatel-cog:
use fstat when fp passed into create media.
Pass in fp and media size to create_multipart.

### DIFF
--- a/cogniac/media.py
+++ b/cogniac/media.py
@@ -295,6 +295,7 @@ class CogniacMedia(object):
             return resp.json()
 
         idx = 1
+        mfp.seek(0)
         while True:
             chunk = mfp.read(chunk_size)
             if not chunk:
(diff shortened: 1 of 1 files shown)


cogniac/media.py
```diff
@@ -216,14 +216,21 @@ def create(cls,
...
+                fsize = stat(filename).st_size
                 # use the multipart interface for large files
-                return CogniacMedia._create_multipart(connection, filename, args)
+            else:
+                fp.seek(0, SEEK_END) # most reliable way of finding file size.
+                fsize = fp.tell()
+                fp.seek(0)
+
+            if fsize > 12 * 1024 * 1024:
+                print("%s using multipart upload" % filename)
```
wskish, line 232: currently sdk is generally nonverbose so prefer to not to add print stuff

wskish (changes requested):

cogniac/media.py
sergii-bond: Amy asks to also remove 
            print(args)
            print(files)
            
can we do it in this pull request for economy of time ?

sergii-bond:
@spatel-cog 
Amy asks to also remove
print(args)
print(files)

are these lines useful ?

PR #99, opened 2021-07-26
Move API version out of URL prefix in core requests wrapper to enable per-endpoint versioning.

mgub:
The intent here is to enable per-endpoint API versioning so service APIs can be introduced incrementally (e.g., pending services with API version `/21/`).

No endpoints were introduced with new versions. This PR just places the version for APIs in endpoint calls.

These changes are being made such that the `COG_URL_PREFIX` environment variable no longer requires the API version in the string. For example, where the API base URL and version were previously set (e.g., `export COG_URL_PREFIX=https://api.cogniac.io/1/`), now only the base API URL would need to be set (e.g., `export COG_URL_PREFIX=https://api.cogniac.io/`). The SDK will detect if the version is specified and strip it from the `COG_URL_PREFIX` if present.

If `COG_URL_PREFIX` includes the version, it will be stripped.

### Summary of changes

- [x] Added endpoint version for following occurrences:
    - [x] `requests.get`, `requests.post`, `requests.delete`
    - [x] `_get(`
    - [x] `_post(`
    - [x] `_delete(`
    - [x] `"url ="`
- [x] test SDK with `COG_URL_PREFIX=https://api.cogniac.io/1` (backward compatibility)
- [x] test SDK  with `COG_URL_PREFIX=https://api.cogniac.io/1/` (backward compatibility)
- [x] test SDK  with `COG_URL_PREFIX=https://api.cogniac.io`
- [x] test SDK  with `COG_URL_PREFIX=https://api.cogniac.io/`

### DIFF
--- a/cogniac/app.py
+++ b/cogniac/app.py
@@ -202,7 +202,7 @@ class CogniacApplication(object):
         Return the integer number of feedback requests pending for this application.
         This is useful for controlling the flow of images input into the system to avoid creating too many backlogged feedback requests.
         """
-        resp = self._cc._get("/21/applications/%s/feedback/pending" % self.application_id)
+        resp = self._cc._get("/1/applications/%s/feedback/pending" % self.application_id)
         return resp.json()['pending']
 
     ##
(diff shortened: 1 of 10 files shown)


sergii-bond (commented):
does it need modification inside _get and _post functions ? 
(after submitting I saw it is work in progress)

mgub:
@sergii-bond NP. Yes, it will need those modifications.

sergii-bond (approved):

wskish (approved):

PR #117, opened 2021-12-08
Support cloudflow

spatel-cog:
Address support for cloudflow /1/process calls.
The following will now work for cloudflow and local (on-prem) edgeflow systems.
```
ef = cc.get_edgeflow('xxxx').  # xxx is the edgeflow_id or cloudflow_id
ef.process_media(subject_uid, media_filename)
```

### DIFF
--- a/cogniac/edgeflow.py
+++ b/cogniac/edgeflow.py
@@ -5,20 +5,21 @@ Copyright (C) 2016 Cogniac Corporation
 """
 
 import os
-import six
-import sys
-import requests
 import re
+from time import time
+import six
+import requests
 from retrying import retry
 from requests.packages.urllib3 import Retry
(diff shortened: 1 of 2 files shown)


cogniac/edgeflow.py
```diff
@@ -5,22 +5,25 @@
...
 from retrying import retry
 from requests.packages.urllib3 import Retry
 from requests.adapters import HTTPAdapter
-from time import time
 
 from .common import server_error, raise_errors
-
 from .media import file_creation_time
 
+from pprint import pprint
```
wskish, line 19: is this used?

spatel-cog: left over from some testing - I will remove.

wskish (approved):

PR #121, opened 2022-01-28
Support non-versioned API endpoint requests.

ah-cog:
Do we want to support this kind of optional behavior so calls to `/efops` can be made (and other non-versioned endpoints)?

Mostly opening to present this as a possible solution.


ah-cog:
Stale.

PR #125, opened 2022-03-21
Adding cumulative reporting of stats for a time window

rahulmishra0802:


cogniac/edgeflow.py
```diff
@@ -350,3 +350,38 @@ def get_next(url):
                 if limit and count == limit:
                     return
             url = resp['paging'].get('next')
+
+
+    def get_xput_stats(self, last_x_minutes=5):
```
wskish, line 355: probably should use start_time and end_time here to make this more closely match the API.  can default the values to last five minutes if not specified by user

wskish (changes requested):

cogniac/edgeflow.py
```diff
@@ -350,3 +350,38 @@ def get_next(url):
...
+        start_time = max(0, curr_time - int(last_x_minutes)*60)
+        status = self.status(subsystem_name='model_detections*',
+                             start=start_time,
+                             end=curr_time)
+
+        model_detections = 0
+        xput_media_pixels = 0
+        xput_gpu_pixels = 0
+        for s in status:
+            tenant_id = s['subsystem'].split('_')[2]
```
spatel-cog, line 371: this is the app_id not tenant_id

cogniac/edgeflow.py
```diff
@@ -350,3 +350,38 @@ def get_next(url):
...
+            status = s['status'].get(tenant_id)
```
spatel-cog, line 372: can you use a different variable name here rather than status as it's already used a place holder for the generator at the start of the for-loop

cogniac/edgeflow.py
```diff
@@ -350,3 +350,38 @@ def get_next(url):
...
+            model_detections += status.get('model_detections', 0)
+            media_pixels = status.get('xput_media_pixels', 0)
+            gpu_pixels = status.get('xput_gpu_pixels', 0)
+
+            if gpu_pixels:
+                xput_media_pixels += media_pixels
+                xput_gpu_pixels += gpu_pixels
+
+        xput_stats = {
+                        'minutes': int((curr_time - start_time)/60),
```
spatel-cog, line 382: seconds might be better here.

cogniac/edgeflow.py
```diff
@@ -350,3 +350,38 @@ def get_next(url):
...
+                        'model_detections': model_detections,
+                        'xput_media_pixels': xput_media_pixels,
```
spatel-cog, line 384: besides the total tput, do you want to provide a breakdown on a per-app basis?

cogniac/edgeflow.py
rahulmishra0802: We can do that. But does someone actually provide time in the timestamp format? It's not very intuitive.

cogniac/edgeflow.py
rahulmishra0802: Will correct

cogniac/edgeflow.py
rahulmishra0802: sure

cogniac/edgeflow.py
rahulmishra0802: Agreed. Seconds would be better

cogniac/edgeflow.py
rahulmishra0802: Sure, should we go for per-app basis and total tput OR just the per-app basis? 
@spatel-cog @wskish

spatel-cog: I think per-app and total tput for during this measurement interval would be useful - the replicas for a given app would have to be accounted for.

rahulmishra0802: I think replicas would only be reflected in GPU_IDs. So code would handle that fine.

cogniac/edgeflow.py
```diff
@@ -350,3 +350,49 @@ def get_next(url):
...
+                         timestamp > start (seconds since epoch)
+        end (float)      filter by last update
+                         timestamp < end   (seconds since epoch)
+        """
+        if end is None:
+            end = time()
+        if start is None:
+            start = end - 300
+        events = self.status(subsystem_name='model_detections*',
+                             start=start,
```
wskish, line 370: this filters based on cloudtime timestamp, which can be delayed relative to actual edgeflow timestamp by some arbitrary time if the connection to cloudcore is slow or temporarily disconnected

cogniac/edgeflow.py
```diff
@@ -350,3 +350,49 @@ def get_next(url):
...
+                             end=end)
+        aggregated_stats = {'total':{}, 'app':{}}
+        total_model_detections = 0
+        total_aggregated_media_pixels = 0
+        total_aggregated_gpu_pixels = 0
+        for event in events:
```
wskish, line 376: add additional filter based on edgeflow timestamp

wskish (changes requested):

cogniac/edgeflow.py
rahulmishra0802: In that case, we might have to fetch all the event reports that has happened after the start timestamp on that EF. This will make the code super slow.

cogniac/edgeflow.py
rahulmishra0802: sure, I will add his filter.

setup.py
```diff
@@ -9,5 +9,5 @@
       author = 'Cogniac Corporation',
```
rahulmishra0802, line 9: Should I update the version? 
If yes, should it be `2.0.14`?

rahulmishra0802:
I have created the PR for public-api changes. https://github.com/Cogniac/public-api/pull/3361 
We need to wait for it to be deployed before we merge this.

wskish (approved):

PR #140, opened 2026-04-02
Add cog user command for system roles

wskish:
## Summary
- Adds `cog user` command that calls `GET /1/users/current` and displays the full user object including system roles
- System roles (`cogniac_admin`, `cogniac_dev`, `ef_ops`, etc.) were previously not visible via any CLI command — `cog tenants` only shows per-tenant roles

## Test plan
- [x] `COG_TENANT=<id> cog user` returns user info with `roles` array
- [x] Works with both JSON (default) and table format

### DIFF
--- a/cogniac/cli.py
+++ b/cogniac/cli.py
@@ -367,20 +367,6 @@ def cmd_auth(args):
 
 def cmd_user(args):
     """Show current user info including system roles."""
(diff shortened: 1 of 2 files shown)


ah-cog:
@wskish Should work now.

spatel-cog (approved):

PR #143, opened 2026-04-09
Fix url_prefix being overwritten by tenant region

### LINKED ISSUE #92: url_prefix is overwritten
cc = cogniac.CogniacConnection(url_prefix=<some url_prefix>)

it is then not used, and overwritten by 
https://github.com/Cogniac/cogniac-sdk-py/blob/ea692495c2534d35374f43eaa1682eba0abe3706/cogniac/cogniac.py#L169

wskish:
## Summary
- Honor explicitly-provided `url_prefix` parameter and `COG_URL_PREFIX` env var instead of silently overwriting them with the tenant's region
- Bump version to 2.0.20

Fixes #92

## Test plan
- [ ] Verify `CogniacConnection(url_prefix="https://staging.cogniac.io")` retains the staging URL after connection
- [ ] Verify default behavior (no explicit url_prefix) still applies tenant region as before

### DIFF
--- a/cogniac/cogniac.py
+++ b/cogniac/cogniac.py
@@ -161,8 +161,8 @@ class CogniacConnection(object):
         # get tenant and user objects associated with this connection
         if self.tenant_id is not None:
             self._tenant = CogniacTenant.get(self)
-            if self._tenant.region is not None:
-                # use tenant object's specified region preference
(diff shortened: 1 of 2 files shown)


wskish:
The fix preserves the region redirect for the normal case while respecting explicit overrides.

ah-cog (approved):

Issue #18, opened 2018-01-31
create media times out while public-api/media-service is processing media


pmranade:
The default timeout for the connection is set to 20sec
This is smaller than the timeout enforced by by ELB or nginx (60 sec by default)
It is good to keep this short as this is session timeout for all calls but we might want to increase this at least for the post media calls as those operations can take longer than 20 sec

pmranade:
I tried changing all the timeouts in media.py to be 20 times the timeout in cogniac.py (just to make it large, we can figure out what this no should really be or fix the timeou at 60 sec for media calls)
I tried uploading 900+Mb file 5 times and was able to upload successfully.
When I removed the timeout multiplier from media.py and tried multiple times, I got the timeout error everytime

wskish:
closing as the default timeout is 60s now.  https://github.com/Cogniac/cogniac-sdk-py/blob/60cb735a5f7f7346bf5ff94c03cceef7d8091392/cogniac/cogniac.py#L76

PR #144, opened 2026-04-11
Add async SDK and modernize dependencies

wskish:
## Summary

Adds a full async/await API to the SDK alongside the existing sync API, and modernizes the dependency stack.

### Dependency modernization
- **Drop Python 2.7** — remove `six`, set `python_requires>=3.11`
- **`requests` → `httpx`** — supports both sync (`httpx.Client`) and async (`httpx.AsyncClient`) from one library
- **`retrying` → `tenacity`** — works transparently on both sync and async functions, same retry semantics

### Async API
Every sync entity class has an async counterpart:

| Sync | Async |
|------|-------|
| `CogniacConnection(...)` | `await AsyncCogniacConnection.create(...)` |
| `CogniacSubject` | `AsyncCogniacSubject` |
| `CogniacApplication` | `AsyncCogniacApplication` |
| `CogniacMedia` | `AsyncCogniacMedia` |
| `CogniacTenant` | `AsyncCogniacTenant` |
| `CogniacUser` | `AsyncCogniacUser` |
| `CogniacEdgeFlow` | `AsyncCogniacEdgeFlow` |
| `CogniacNetworkCamera` | `AsyncCogniacNetworkCamera` |
| `CogniacExternalResult` | `AsyncCogniacExternalResult` |
| `CogniacOpsReview` | `AsyncCogniacOpsReview` |

Key design choices:
- **Factory pattern** for async connection — `await AsyncCogniacConnection.create()` since `__init__` can't be async
- **Context manager** — `async with await AsyncCogniacConnection.create() as cc:` for automatic cleanup
- **Explicit `set()` method** — `await subject.set(name="new", description="updated")` instead of `__setattr__` magic (also batches multiple fields into one API call)
- **Async generators** for paginated endpoints (`detections`, `media_associations`, `usage`, etc.)
- **Sync API unchanged** — existing code continues to work identically

### Test suite
Added pytest-based integration tests against a live Cogniac tenant (42 passed, 1 skipped):
- 26 sync baseline tests (connection, tenant, user, subjects, apps, media, edgeflows, associations)
- 13 async tests (connection, subjects, apps, media lifecycle)
- 3 sync/async parity tests (verify both APIs return identical results)

### Other fixes
- `cogupload` script updated to use tenacity and Python 3 print()
- README rewritten with standard markdown formatting
- CLAUDE.md updated to document dual sync/async architecture

## Test plan
- [x] All 26 sync baseline tests pass (no regressions from dep swap)
- [x] All 13 async integration tests pass
- [x] All 3 parity tests confirm sync/async equivalence
- [x] `cogupload` script updated and syntax-checked


spatel-cog (commented):
## Code Review: Add async SDK and modernize dependencies

Good work — the dep modernization (requests→httpx, retrying→tenacity, drop six/Py2) is clean and the async API design choices are sound. The factory pattern for `AsyncCogniacConnection.create()` and the explicit `set()` method are the right call. The test suite with parity tests is a nice touch.

That said, there are several bugs in the async media path and some design gaps that should be addressed before merge. Organized by severity:

---

### Bugs — must fix

**1. File handle leaks in `async_media.py` `create()` and `_create_multipart()`**

`create()` line 261 opens a file but never closes it if the POST raises:
```python
files = {filename: open(filename, 'rb')}   # leaked on exception
```

`_create_multipart()` line 288 also leaks:
```python
mfp = open(filename, 'rb')   # never closed after upload loop or on exception
```

Fix: use context managers or try/finally.

**2. `download()` is not actually async — buffers entire response in memory**

Line 367:
```python
for chunk in resp.iter_bytes(chunk_size=512 * 1024):
```

`_get()` returns a fully-read `httpx.Response`. By the time `iter_bytes()` runs, the entire body is already in memory. For large video files, this defeats the purpose of chunked writing.

For true streaming, you'd need a `_get_stream()` method on the connection that uses `session.stream('GET', url)` and `async for chunk in resp.aiter_bytes()`.

**3. `download()` closes the caller's file handle (line 370)**

```python
filep.close()
```

The caller opens the file, the callee closes it. This violates resource ownership conventions and will cause `ValueError: I/O operation on closed file` if the caller tries to use the handle afterwards.

**4. Blocking I/O in async functions throughout `async_media.py`**

These calls block the event loop:
- Line 244: `stat(filename).st_size`
- Lines 276-285: `md5_hexdigest()` — reads entire file in 8MB sync chunks. On a 100MB video, this blocks the event loop for hundreds of ms.
- Line 288: `open(filename, 'rb')`
- Lines 318-321: `mfp.read(chunk_size)` in a while loop

For a library that provides an async API, these should use `asyncio.to_thread()`. At minimum, wrap the MD5 computation.

**5. `AsyncCogniacEdgeFlow.__setattr__` is dead code — no way to persist changes**

Lines 70-75 — both branches of the if/else do the same thing:
```python
def __setattr__(self, name, value):
    if name not in self._edgeflow_keys:
        super().__setattr__(name, value)
        return
    super().__setattr__(name, value)  # identical to above
```

And there's **no `set()` method** on this class. There is literally no way to persist attribute changes to the server for EdgeFlows via the async API.

**6. `requirements.txt` is stale — lists old dependencies**

Still has `retrying`, `requests`, `six`. `setup.py` correctly has `httpx`, `tenacity`, `tabulate`. Anyone doing `pip install -r requirements.txt` gets the wrong packages. This will break.

**7. `common.py` — `response.content` is bytes, not str**

Lines 49, 53, 57:
```python
msg = "ServerError (%d): %s" % (response.status_code, response.content)
```
In httpx, `response.content` is `bytes`. This produces `ServerError (500): b'{"error": ...}'`. Use `response.text` instead.

**8. `setup.py` — no version lower-bounds on deps**

```python
install_requires=['httpx', 'tenacity', 'tabulate']
```

- `tenacity` **must** be `>=8.0.0` for proper async support. Older versions silently break all retry-decorated async methods.
- `httpx` should be `>=0.24.0` for `follow_redirects` parameter support.

Without these bounds, `pip install cogniac` on a system with old cached versions will silently produce broken behavior.

---

### Design issues — should fix before merge

**9. Silent local-only writes are a migration trap**

In the sync API, `subject.name = "new"` auto-persists to the server via `__setattr__`. In the async API, the same code silently stores the value locally only. Users migrating from sync will be bitten.

Recommendation: make `__setattr__` raise for mutable server-managed keys:
```python
def __setattr__(self, name, value):
    if name in self.mutable_keys:
        raise AttributeError("Use 'await obj.set(%s=...)' to update" % name)
    ...
```

**10. `set()` key validation inconsistent across classes**

- `AsyncCogniacSubject.set()` and `AsyncCogniacMedia.set()`: validate both mutable and immutable keys
- `AsyncCogniacTenant.set()`: no validation at all
- `AsyncCogniacEdgeFlow`: no `set()` method at all

**11. Async `get_all` methods use `connection.tenant_id` directly**

Sync uses `connection.tenant.tenant_id` which triggers the `_require_tenant()` property and gives a helpful error listing available tenants. Async uses `connection.tenant_id` directly which will be `None` if unset, producing a confusing 404 from `/1/tenants/None/...`.

**12. Pagination KeyError risk**

All async generators do `resp['paging'].get('next')`. If the API ever omits the `paging` key, this raises `KeyError`. Safer: `resp.get('paging', {}).get('next')`.

**13. Mutable default argument in `search()`**

`async_subject.py` line 108:
```python
async def search(cls, connection, ids=[], ...):
```
Classic Python gotcha. Use `ids=None` and default in the body.

**14. `AsyncCogniacConnection.create()` docstring is wrong (line 34)**

```python
async with AsyncCogniacConnection.create(...) as cc:
```
Missing `await`. Correct pattern (which the tests get right):
```python
async with await AsyncCogniacConnection.create(...) as cc:
```

---

### Missing async methods vs sync

| Sync method | Status in async |
|-------------|-----------------|
| `CogniacSubject.create_reference_media()` | Missing |
| `CogniacApplication.add_input/output_subject()` | Missing |
| `CogniacApplication.model_name()` / `download_model()` | Missing |
| `CogniacApplication._CogniacAppTypeConfig` (auto-persist) | Simplified to plain dict |
| `CogniacEdgeFlow` attribute persistence | No `set()`, no `__setattr__` persist |
| `CogniacConnection.get_all_authorized_tenants()` | No async equivalent |
| `CogniacConnection` convenience methods (get_all_subjects, etc.) | By design — users call entity classes directly |

If intentionally deferred, document what's missing in the README.

---

### Sync side — pre-existing issues to fix while here

**15. Bare `except:` in `cogniac.py` lines 64, 124, 143** — catches `KeyboardInterrupt`/`SystemExit`. Use `except KeyError:`.

**16. Stray `print(resp.json())` in `external_results.py` line 105** — debug statement in production code. The async version correctly removed it; sync should too.

**17. `get_version()` double-calls `raise_errors()`** — in both `cogniac.py:519` and `async_connection.py:294`. Already called inside `_get()`.

**18. CLI `cmd_user`, `cmd_deployments_*`, `cmd_workflows_get` bypass `_get()` retry/re-auth** — use raw `cc.session.get()` instead.

---

### Test coverage gaps

- **No test for `set()` on any entity** — this is the headline new async API
- **No test for multipart upload** (>12MB triggers `_create_multipart`)
- **No test for async generators** (`media_associations`, `status`, ops_review `search`)
- **Zero async tests for:** Tenant, User, EdgeFlow, ExternalResult, OpsReview, NetworkCamera (5 of 9 async entity classes)
- **No session-scoped async fixture** — every async test creates its own connection (extra auth round-trips)
- **JPEG bytes duplicated 3x** across test files — extract to conftest fixture
- **`pytest-asyncio` and `pytest-timeout` not declared** in any requirements/dependency file

---

### Summary

The architecture is right and the dep migration is correct. The main concerns are:
1. **`async_media.py`**: file handle leaks, blocking I/O, non-streaming download
2. **`__setattr__` migration trap**: silent local-only writes across all async entities
3. **`requirements.txt` stale** and **no dep version bounds** in setup.py
4. **`common.py` bytes vs text** in error messages

Fix those and this is ready.

| Severity | Count |
|----------|-------|
| Bugs (must fix) | 8 |
| Design (should fix) | 6 |
| Missing methods (document) | 7 |
| Sync cleanup | 4 |
| Test gaps | 7 |

spatel-cog:
## Suggestion: modernize project tooling while deps are changing

Since this PR is already overhauling dependencies (requests to httpx, retrying to tenacity, dropping six), it would be a natural time to migrate from `setup.py` + `requirements.txt` to `pyproject.toml` + `uv`. This would fix the stale `requirements.txt` bug and modernize the build in one shot.

### What this looks like

Replace `setup.py` and `requirements.txt` with a single `pyproject.toml`:

```toml
[project]
name = "cogniac"
version = "3.0.0"
description = "Python SDK for Cogniac Public API"
requires-python = ">=3.11"
dependencies = [
    "httpx>=0.24.0",
    "tenacity>=8.0.0",
    "tabulate",
]

[dependency-groups]
dev = [
    "pytest",
    "pytest-asyncio",
    "pytest-timeout",
]

[project.scripts]
cogniac = "cogniac.cli:main"

[project.urls]
Homepage = "https://github.com/Cogniac/cogniac-sdk-py"
```

Then `uv lock` generates a lockfile with pinned transitive deps (replaces `requirements.txt`), and `uv sync` installs everything reproducibly.

### Why this helps

1. **Fixes the stale requirements.txt** -- the lockfile is always in sync with `pyproject.toml` because `uv lock` regenerates it
2. **Declares test deps properly** -- `pytest-asyncio` and `pytest-timeout` are not listed anywhere today; `[dependency-groups]` solves this
3. **Enforces version lower-bounds** -- `tenacity>=8.0.0` and `httpx>=0.24.0` live in one place
4. **Faster iteration** -- `uv pip install` is 10-50x faster than pip, noticeable when running the test suite repeatedly
5. **setup.py is legacy** -- pyproject.toml is the standard since PEP 621

This is optional and could be a follow-up PR if you want to keep this one focused on the async API.

wskish:
## Review Response — 443216e

Addressed all issues from @spatel-cog's review (plus a bonus bug fix). **66 tests pass, 7 skipped** (no edgeflows/cameras on test tenant).

### Bugs fixed (9)
| # | Issue | Fix |
|---|-------|-----|
| 1 | File handle leaks in `async_media.py` | `with open()` in `create()`, `try/finally` in `_create_multipart()` |
| 2 | `download()` not truly async | Added `_stream()` to connection + `aiter_bytes()` |
| 3 | `download()` closes caller's file handle | Removed `filep.close()` |
| 4 | Blocking I/O in async media | Wrapped `stat()`, `file_creation_time()`, `md5_hexdigest()`, chunk reads in `asyncio.to_thread()` |
| 5 | `AsyncCogniacEdgeFlow` dead `__setattr__` + no `set()` | Fixed guard, added `set()` method |
| 6 | `requirements.txt` stale | Replaced with `httpx>=0.24.0`, `tenacity>=8.0.0`, `tabulate` |
| 7 | `common.py` `response.content` (bytes) in errors | → `response.text` |
| 8 | No dep version bounds in `setup.py` | `httpx>=0.24.0`, `tenacity>=8.0.0` |
| **bonus** | `ExternalResult.search()` passes `json=` to GET | → `params=` (pre-existing bug from requests→httpx migration, both sync+async) |

### Design fixes (8)
| # | Issue | Fix |
|---|-------|-----|
| 9 | Silent local-only writes (migration trap) | `__setattr__` raises on mutable keys across **all 7 async entities**, directing to `set()` |
| 10 | `set()` validation missing on `AsyncCogniacTenant` | Added mutable/immutable key validation |
| 11 | `get_version()` double `raise_errors()` | Removed redundant call (sync + async) |
| 12 | Bare `except:` in `cogniac.py` | → `except KeyError:` (3 locations) |
| 13 | Stray `print(resp.json())` in `external_results.py` | Removed |
| 14 | Docstring missing `await` | Fixed context manager example |
| 15 | Pagination `KeyError` risk | `resp['paging']` → `resp.get('paging', {})` across 5 files |
| 16 | Mutable default `ids=[]` | → `ids=None` with body defaulting |

### Test coverage added (24 new tests)
- **New async entity tests**: Tenant (4), User (3), EdgeFlow (4), NetworkCamera (4), ExternalResult (1), OpsReview (2)
- **`__setattr__` guard tests** on Subject, App, Media, Tenant, User, EdgeFlow, Camera
- **Download-to-file test** exercising `_stream()` + `aiter_bytes()`
- **Parity tests** for tenant and edgeflows
- Extracted JPEG bytes to `conftest.py` shared fixture
- Added `requirements-test.txt` (pytest-asyncio, pytest-timeout)

### Deferred
- **`set()` integration tests** — deferred until preferred test tenant identified
- **Multipart upload test** (>12MB) — deferred (expensive for CI)
- **Missing async methods** (create_reference_media, add_input/output_subject, model ops) — separate PR
- **CLI bypassing `_get()` retry/re-auth** (#18) — pre-existing, out of scope

wskish:
tooling update incoming
skipping write tests  until have better test tenant strategy

wskish:
@spatel-cog Implemented your pyproject.toml + uv suggestion in b3ee553:

- `setup.py`, `requirements.txt`, `requirements-test.txt` → single `pyproject.toml`
- `uv.lock` provides reproducible pinned transitive deps
- Test deps declared in `[dependency-groups] dev`
- All CLI entry points preserved (`cogniac` via `[project.scripts]`, legacy bin/ scripts via `[tool.setuptools] script-files`)
- 66 tests pass, 7 skipped — no regressions

Ready for review.

spatel-cog (approved):

spatel-cog (approved):

PR #147, opened 2026-05-04
subject: allow consensus=True/False with associate_media()

bjw-0:
## What

Removes an overly strict assertion in `CogniacSubject.associate_media()` (and its async counterpart) that raised `AssertionError` when callers passed `probability` alongside an explicit `consensus` value.

## Why

`consensus='True'` or `'False'` associations don't use `probability` — it's irrelevant when consensus is explicitly set. The old behavior forced callers working with consensus GT data (e.g. segmentation apps that need `consensus=True, focus=box`) to drop down to raw HTTP calls:

```python
# Workaround that should no longer be necessary:
cc.session.post(".../1/subjects/%s/media" % uid, json={
    "media_id": mid, "consensus": "True", "focus": {"box": box}, ...
})
```

With this fix, `associate_media(media, focus=focus, consensus='True', ...)` works correctly whether or not `probability` is also passed — `probability` is simply ignored when consensus is not `'None'`.

The fix is identical in `subject.py` and `async_subject.py`.

## Test

```python
subject.associate_media(media, focus={"box": {...}}, consensus='True')  # no longer raises
subject.associate_media(media, focus={"box": {...}}, consensus='False') # no longer raises
subject.associate_media(media, probability=0.9)                         # unchanged
```

Fixes Cogniac/cogniac-kb#38

cc @wskish

### DIFF
--- a/cogniac/async_subject.py
+++ b/cogniac/async_subject.py
@@ -256,8 +256,7 @@ class AsyncCogniacSubject(object):
         if consensus == 'None' and probability is None:
             data['uncal_prob'] = .99
 
-        if probability is not None:
-            assert(consensus == 'None')
+        if probability is not None and consensus == 'None':
             data['uncal_prob'] = probability
 
         if enable_wait_result:
--- a/cogniac/subject.py
+++ b/cogniac/subject.py
@@ -401,8 +401,7 @@ class CogniacSubject(object):
         if consensus == 'None' and probability is None:
             data['uncal_prob'] = .99
 
-        if probability is not None:
-            assert(consensus == 'None')
+        if probability is not None and consensus == 'None':
             data['uncal_prob'] = probability
 
         if enable_wait_result:


wskish (commented):
## Review

### Previous intention of the assertion

The `assert(consensus == 'None')` when `probability is not None` was a **contract guard**, not arbitrary strictness. Per the docstring:

> `probability (float)  association probability, only valid if consensus = 'None'`

Semantics:
- `consensus='None'` → consensus unknown; caller supplies `uncal_prob` (defaults to `0.99` if `probability is None`).
- `consensus='True'/'False'/'Sidelined'` → consensus is decisive; probability is meaningless.

The assertion was telling callers: *"these two arguments are mutually exclusive — if you're passing both, you're confused about which mode you're in."*

### Concerns

1. **Lost safety signal.** The old behavior was "you violated a contract → AssertionError." The new behavior is "we silently drop one of your arguments." That's a regression for callers who genuinely mis-pass both. Consider at minimum a `warnings.warn(...)` (or `logger.warning`) when `probability` is provided but ignored — preserves the unblock while keeping the diagnostic.

2. **The motivating workaround in the PR description doesn't pass `probability`:**
   ```python
   cc.session.post(..., json={"media_id": mid, "consensus": "True", "focus": {"box": box}, ...})
   ```
   With `consensus='True'` and `probability=None` (default), the current assertion does **not** fire — I walked the function on `master` to confirm. So either (a) the workaround shown isn't the literal call that was blocked, or (b) some upstream wrapper is defaulting `probability` to a non-`None` value and that's the real culprit. Worth nailing down — if it's (b), the cleaner fix is at the caller.

3. **Docstring not updated.** Still says probability is "only valid if consensus = 'None'." After this change, the actual behavior is "probability is silently ignored unless consensus = 'None'." Update the docstring so future readers don't rely on the old contract.

4. **Symmetric fix in `subject.py` and `async_subject.py`** — good, no drift.

### Recommendation

Merge-worthy in spirit (the workaround is ugly), but before merging I'd like to:
- Confirm/identify which caller path was actually triggering the assert (point 2),
- Add a `warnings.warn(...)` when `probability` is dropped (point 1),
- Update the docstring (point 3).

bjw-0:
@wskish — your point 2 is sharper than my PR body acknowledged. Walked the function on `main` myself just now:

```python
if consensus == 'None' and probability is None:
    data['uncal_prob'] = .99
if probability is not None:
    assert(consensus == 'None')
    data['uncal_prob'] = probability
```

With the documented reproducer from `Cogniac/cogniac-kb#38`:

```python
subject.associate_media(media, focus=focus, consensus='True', ...)  # probability defaults to None
```

`probability is not None` evaluates False — assertion never fires. So the example I cited in the PR body **cannot** be what triggered the AssertionError reported in cogniac-kb#38. Two possibilities:

1. **The original report misattributed the error.** The reporter (past-me, in a prior session) hit AssertionError somewhere — possibly the *other* assert in this function, `assert(consensus in ['True', 'False', 'None', 'Sidelined'])` at line 401, fired because the caller passed `consensus=True` (boolean) instead of `'True'` (string). That would have looked the same in a traceback. The real fix would be a more informative error message at line 401, or accepting both forms.
2. **The caller was wrapping the call** somewhere and defaulting `probability` to a non-None value (e.g. `probability=0.99` via a helper) — but I just grepped agent-ops and cogniac-kb and there's no such wrapper.

(1) is the much more likely explanation. I'm going to:

- **Close this PR** unless we can identify a concrete caller path that the current code actually rejects.
- **Re-open kb#38 with a corrected reproducer** that captures whichever path was actually broken (likely the boolean-vs-string `consensus` arg, or the separate feedback-endpoint focus-stripping issue, which is point 2 of kb#38 and is the real architectural problem).
- **Apply your points 1, 3, 4 as good general feedback** regardless: if/when we touch this function, the `warnings.warn` on silent-drop pattern and the docstring update are the right shape.

Going to close this in 24h unless someone surfaces the actual blocked caller. Thanks for the careful read — saved a confused merge.

bjw-0:
Closing per the 2026-05-11 24h commitment. No one surfaced an actual blocked caller path during the window, and walking the function on `main` confirms the assertion at line 408 (`if probability is not None: assert(consensus == 'None')`) doesn't fire on the documented kb#38 reproducer (`probability=None` default → second `if` doesn't trigger).

The likely root cause for whatever AssertionError the kb#38 reporter hit was the **other** assert in the function — `assert(consensus in ['True', 'False', 'None', 'Sidelined'])` at line 401 — firing because the caller passed Python `True` (bool) instead of `'True'` (string). Same traceback shape, easy misattribution.

If a real blocked caller surfaces later, two cleaner shapes than this PR:

1. **Better error at line 401** — replace bare `assert` with `ValueError("consensus must be one of '...', got %r — note: must be a string, not a Python bool" % consensus)`. Catches the bool-vs-string case explicitly.
2. **`warnings.warn` on silent-drop** — if we ever do want to allow `probability` alongside `consensus='True'/'False'`, emit a runtime warning rather than silently dropping. (Per wskish's point 1 in his original review.)

Either should be a fresh PR with a corrected reproducer. Reopen this branch if I'm wrong about the root cause and want to revisit.

bjw-0:
Closing per the self-imposed deadline in my [2026-05-11 comment](https://github.com/Cogniac/cogniac-sdk-py/pull/147#issuecomment-4427) — no one surfaced a concrete caller path that the current code actually rejects, and walking the function on `master` confirmed the example I originally cited can't trigger the `assert(consensus == 'None')`.

Reasoning recap (full version in the earlier comment):
- With `consensus='True'` and default `probability=None`, `probability is not None` evaluates `False`, so the assertion never fires.
- The `AssertionError` originally attributed to this code path in [cogniac-kb#38](https://github.com/Cogniac/cogniac-kb/issues/38) was most likely the *other* assert at line 401 (`assert(consensus in ['True', 'False', 'None', 'Sidelined'])`) firing on a boolean-vs-string `consensus` arg, or the separate feedback-endpoint focus-stripping issue (point 2 of kb#38).
- @wskish's general feedback (warn-on-drop, docstring update, symmetric fix) still stands as good shape if/when we touch this function for a real bug fix — just not on this PR.

If anyone hits the original AssertionError under a different reproducer, please re-open with that reproducer attached and I'll re-target the fix.

PR #154, opened 2026-05-22
Fix post_feedback dropping per-ROI focus; bump to 3.0.5

wskish:
## Summary

- Add `focus=None` kwarg to `CogniacApplication.post_feedback` (and the async counterpart) and forward it at the top level of the `/21/applications/<id>/feedback` payload.
- Document that a `focus` key placed inside a subject dict is ignored server-side — the trap that made the previous failure silent.
- Bump to 3.0.5.

Fixes #153.

## Test plan

- [ ] Against a box_detection app: `app.post_feedback(mid, [...], focus={'box': {...}})` results in an output-subject association with `focus={'box': {...}}` (not `None`).
- [ ] Calling without `focus` continues to work for apps that don't need it.
- [ ] Async path (`AsyncCogniacApplication.post_feedback`) behaves the same.

### DIFF
--- a/cogniac/app.py
+++ b/cogniac/app.py
@@ -225,7 +225,7 @@ class CogniacApplication(object):
     #  post_feedback
     ##
     @retry(stop=stop_after_attempt(8), wait=wait_exponential(multiplier=0.5), retry=retry_if_exception(server_error))
-    def post_feedback(self, media_id, subjects):
+    def post_feedback(self, media_id, subjects, focus=None):
         """
         Provides feedback to the application for a given subject-media assocation; returns None.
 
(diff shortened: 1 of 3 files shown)


CogniacViSean (approved):
**Tested with CN tenant (SDK 3.0.5 from this branch).**

- [x] **Sync path with focus** — `app.post_feedback(mid, [subj], focus={'box': {'x0': 7321, 'x1': 7568, 'y0': 382, 'y1': 627}})` on a `box_detection` app (`shndyixf`, Detect hole). Output-subject association confirmed `consensus=True` and `focus={'box': {'y1': 627, 'y0': 382, 'x0': 7321, 'x1': 7568}}` — not `None`. Prior to this fix the same call via `cc._post()` workaround was required; the SDK wrapper silently dropped the focus.

- [x] **No-focus path** — `app.post_feedback(mid, [subj])` (no kwarg) and `app.post_feedback(mid, [subj], focus=None)` (explicit `None`) both returned without error on a second `box_detection` app (`yswyj7wm`, Detect carrier plate BU). Association confirmed `consensus=True`.

- [x] **Async path with focus** — `await async_app.post_feedback(mid, [subj], focus={'box': {...}})` on the same `shndyixf` app. Output-subject association confirmed `consensus=True`, `focus` present with correct box coordinates, and `app_data_type=box_set` / `app_data` intact.

LGTM. The `if focus is not None` guard (vs `if focus`) is the right choice — it avoids accidentally suppressing an empty-dict focus while staying backward-compatible with callers that omit the kwarg entirely.

PR #156, opened 2026-06-10
Add SDK and CLI coverage for additional public API endpoints

ah-cog:
## Summary

Expands the `cogniac` SDK and CLI to cover public API endpoints that previously had no client surface, and restructures the CLI into a consistent **nested, noun-first / verb-last** command tree. JSON output by default; complete result sets returned by default (pagination is followed automatically).

Resource ids are passed as **`--<resource>-id` flags** (e.g. `--application-id`, `--subject-uid`) — the canonical form. For backward compatibility the older **positional** spelling (`application get <id>`) is still accepted as a deprecated form, and prior flat/hyphenated command spellings keep working as hidden aliases — so nothing that worked before stops working.

Four parts below: **(A)** new/changed SDK methods, **(B)** the full CLI command tree, **(C)** pagination, **(D)** what changed in response to review. Please review by section.

## Part A — SDK methods

New methods are added on both the sync entity classes and their async counterparts, and new entity classes are exported from the package root.

- **`CogniacApplication`** — application types (list/get), `classify`, `donate_model`, `export_model_to_meraki`, replay (`replay_status`/`replay_start`/`replay_stop`), `detections_pending`, `event_types`, `events`, `consensus_history`, performance (`performance_current_validation`/`performance_release_validation`/`performance_new_random`), `model_performance`, push (`push_notifications`/`subscribe_push`), feedback (`feedback`, `feedback_request`, `submit_feedback`, `feedback_request_count`, `pending_feedback_requests`, `purge_feedback`, `delete_feedback_requests`), evaluation-metrics (`evaluation_metrics`, `create_evaluation_metric`, `register_default_evaluation_metric`, `copy_evaluation_metrics`), consensus releases (`consensus_releases`, `consensus_release`, `consensus_release_items`, `consensus_release_upstream_assertions`, `consensus_detection_release`), labeling models (`labeling_image_encoder`, `labeling_mask_decoder`), `download_model`, `update`.
- **`CogniacSubject`** — `detections`, `consensus_history`, `disassociate_media`, `update`.
- **`CogniacMedia`** — `share`, `create_detection`, `detections`, `embeddings`, `update`.
- **`CogniacEdgeFlow`** — `create`, `delete`, certificate (`get_certificate`/`set_certificate`/`replace_certificate`/`delete_certificate`), `metrics`, `metric_names`, `all_metrics`, device-control events (`reboot`, `ping`, `upgrade`, `set_boot_software_version`, `factory_reset`, `flush_upload_queue`, `time_bound_media_upload`, `trigger_camera_capture`), `update`.
- **`CogniacTenant`** — EdgeFlow certificate (`get/set/delete`), `delete_meraki_api_key`, invites, `get_cloudcore_import`, tenant-user management (`add_user`/`delete_user`/`set_user_role`).
- **`CogniacNetworkCamera`** — `genicam`, `upload_genicam`, `update`.
- **`CogniacUser`** — `get_all` (returns `CogniacUser` instances), `get_by_id`, `delete_by_id`, `tenants`, `request_password_reset`, invites, API keys (`api_keys`, `api_key`, `create_api_key`, `delete_api_key`).
- **New entity classes** — `CogniacDeployment`, `CogniacDeploymentCapacityClass`, `CogniacWorkflow`, `CogniacBuild` (each sync + async, re-exported from `cogniac`).
- **`CogniacConnection`** — `edgeflows()` is the preferred name; `gateways()` remains a deprecated alias and now emits a `DeprecationWarning`. A `_put` helper was added (sync + async) so PUTs go through credential re-auth.
- **Fix** — body-bearing GETs route through httpx `.request("GET", ...)` (httpx's `.get()` rejects a request body), so `download_model` (GET `/ccp` + `/ccppkg`) works on both clients.
- **Robustness** — HTTP **429 rate-limit** responses are now retried with the existing exponential backoff (the shared retry predicate treats 429 as transient; `Retry-After` is recorded on the exception); and `media_associations()` **normalizes** `app_data` / `custom_data` that the API sometimes returns as serialized JSON strings, so callers always see dicts/lists.

## Part B — CLI commands (nested tree)

Noun → sub-noun → verb (verb last). Resource ids are `--<resource>-id` flags (canonical); the older positional form (`application get <id>`) still works as a deprecated alias. Payloads/modifiers are flags too. Pagination flags (`--limit`, `--cursor`) and sort/filter flags apply to list commands.

The global flags `--format` (`json`/`table`/`jsonl`) and `--tenant` are accepted **before or after** the command (`cogniac application list --format jsonl`). `--body` accepts inline JSON, `@FILE`, or `-` (stdin). Timestamp filters (`--start`/`--end`) accept epoch seconds or ISO 8601.

### auth
```
auth                    check credentials (and, with --tenant/COG_TENANT, that a session can be minted)
auth login              browser-loopback login; stores a per-user API key   [--no-browser]
auth logout             remove the stored login credential
```

### commands
```
commands                print the whole CLI surface as JSON: canonical noun -> verbs -> args
                        (name, positional?, required?, type, choices, help) — one call, no --help walk
```

### tenant   (aliases: tenants)
```
tenant get                                     current tenant   (bare `cogniac tenant` also works)
tenant list                                    tenants you are authorized for
                                               (bare `cogniac tenants` also lists them)
tenant edgeflow-certificate get|set|delete     tenant-wide EdgeFlow cert   [set: --body]
                                               (aliases: gateway-certificate, edgeflow-cert, …)
tenant meraki-api-key delete
tenant cloudcore-import-key get <cloudcore-import-key>
tenant user list
tenant user add        --email <e> [--role <r>]
tenant user delete     --email <e>
tenant user role set   --email <e> --role <r>
```

### application   (aliases: app, apps, applications)
```
application list
application get|update|delete   --application-id <id>          [update: per-field flags and/or --body]
application create              --body <json>
application leaderboard         --application-id <id>  [--set-assignment --snapshot-type --eval-metrics --top --full]
application classify            --application-id <id> <image-file>
application events              --application-id <id>  [--start --end --limit --cursor --reverse --event-type ...]
application event types         --application-id <id>
application detections pending  --application-id <id>
application replay status|start|stop   --application-id <id>   [status: --timeout; start: --body]
application performance current|release|new-random  --application-id <id>  [--start --end --limit --reverse --duration]
application push get|subscribe  --application-id <id>  [--device-id --app-bundle-id --event-type] [subscribe: --unsubscribe]
application feedback list|count|pending|purge|purge-requests  --application-id <id>   [list: --limit --cursor]
application feedback get         --application-id <id> --feedback-id <id>
application feedback create      --application-id <id> --body <json>
application model performance    --application-id <id> --subject-uid <uid> [--consensus --reverse --probability-lower/upper --limit --cursor --set-assignment]
application model donate         --application-id <id> --source-application-id <id>
application model export         --application-id <id> --target meraki
application model download       --application-id <id> [--model-id]
application model list           --application-id <id> [--start --end --limit --reverse]
application consensus history    --application-id <id> [--start --end --limit --subject-uid]
application consensus release list      --application-id <id>
application consensus release get|items|upstream  --application-id <id> --release-id <id>   [items/upstream: --limit --cursor]
application consensus release detections          --application-id <id>
application evaluation metrics get|create|register-default  --application-id <id>   [create/register-default: --body]
application evaluation metrics copy     --source-application-id <id> --target-application-id <id>
application label-image-encoder-model download   --application-id <id> [--body]
application label-mask-decoder-model download     --application-id <id>
application type list   [--production --deprecated --reverse]
application type get    <application-type>
application build list  [--application-id <id>]
application build get|delete   --application-build-id <id>
application build create        --body <json>
application build names list
application build lint  <filename>
```

### subject   (aliases: subjects)
```
subject list
subject get|update|delete  --subject-uid <uid>             [update: per-field flags and/or --body]
subject create  <name>  [--description --external-id]
subject search  [--prefix --similar --name --ids ... --limit]
subject media   --subject-uid <uid>  [--limit(=100) --consensus --probability-lower/upper --reverse/--no-reverse --full-media]
subject associate|disassociate   --subject-uid <uid> --media-id <id>   [associate: --consensus]
subject consensus history  --subject-uid <uid>  [--start --end --limit]
subject detections  --subject-uid <uid> --media-id <id>
```

### media   (no plural alias — uncountable)
```
media get       --media-id <id>  [--download [FILE]]
media upload    <filename>  [--subject-uid --external-media-id --domain-unit --meta-tags ...]
media update|delete   --media-id <id>          [update: per-field flags and/or --body]
media download  --media-id <id>  [-o OUTPUT]
media search    [--md5 --filename --external-media-id --domain-unit --limit]
media share     --media-id <id>  [--body]
media embeddings get   --media-id <id>  [--model-id --focus]
media detection list|create   --media-id <id>   [list: --limit; create: --body]
```

### edgeflow   (aliases: edgeflows, gateway, gateways)
```
edgeflow list
edgeflow get|update|delete   --edgeflow-id <id>          [update: per-field flags and/or --body]
edgeflow create   [--body]
edgeflow status   --edgeflow-id <id>  [--subsystem --limit(=10)]
edgeflow certificate get|set|replace|delete   --edgeflow-id <id>   [set/replace: --body]
edgeflow metrics list  [--edgeflow-id <id>]
edgeflow metrics names
edgeflow event reboot|ping|factory-reset   --edgeflow-id <id>   [ping: --ping-id]
edgeflow event upgrade|set-boot-software-version   --edgeflow-id <id> --software-version <v>
edgeflow event flush-upload-queue   --edgeflow-id <id>  [--start-time --end-time]
edgeflow event time-bound-media-upload   --edgeflow-id <id> --start-time <t> --end-time <t>
edgeflow event trigger-camera-capture   --edgeflow-id <id> --subject-uid <uid> [--trigger-domain-unit]
```

### camera   (aliases: cameras, network-camera, network-cameras)
```
camera list
camera get|update|delete   --network-camera-id <id>        [update: per-field flags and/or --body]
camera create   --body <json>
camera genicam  --network-camera-id <id>
```

### deployment   (aliases: deployments, deployment-group, deployment-groups)
```
deployment list
deployment get|delete   --deployment-group-id <id>
deployment create   [--body]
deployment edgeflows    --deployment-group-id <id>
deployment history      --deployment-group-id <id>  [--limit --reverse/--no-reverse]
deployment prepull status   --deployment-group-id <id>
deployment prepull start    --deployment-group-id <id> --workflow-id <id>
deployment target workflow set   --deployment-group-id <id> --workflow-id <id>
deployment capacity list
deployment capacity get   --capacity-class-id <id>
```

### workflow   (aliases: workflows)
```
workflow list
workflow get|delete   --workflow-id <id>
workflow create   [--body]
workflow edgeflow deployment-targets list   [--edgeflow-model]
workflow version new   --workflow-id <id> --body <json>
workflow version get   --base-id <id> <version>
```

### user   (aliases: users)
```
user list   [--user-id <id> | --tenant-id <id>]
user get|delete|tenants   --user-id <id>
user password reset   <email>
user api-key list     --user-id <id>
user api-key get|delete   --user-id <id> --api-key-id <id>
user api-key create   --user-id <id> --description <text>
```

**Aliases.** Every command also accepts singular/plural and synonym/abbreviation spellings of each token (the per-section aliases above, plus sub-noun synonyms like `cert`/`certificate`, `eval`/`evaluation`, `assertion`/`detection`, `event`/`events`, `api-key`/`api-keys`) in every position — including compounds, so `tenant gateway-certificate get` and `gateway event reboot` resolve to the same handlers as the `edgeflow` spellings. Previous flat/hyphenated spellings (e.g. `application-feedback list`, `edgeflow-certificate get`) also remain as hidden aliases.

## Part C — Pagination

List commands **return the complete result set by default** — the SDK methods are generators that follow the API's pagination automatically, and the CLI collects them into one JSON array. `--limit N` caps the total; `--cursor` resumes where the endpoint supports it; sort/filter flags still apply. Note: the `feedbackRequests` and `consensus_release` sub-collections return the full set in one response (no server-side limit/cursor) — `--limit` is applied client-side there and `--cursor` is best-effort.

The two **expensive reads keep a default cap** so a bare call can't walk an entire device/subject history: `edgeflow status` defaults to `--limit 10` and `subject media` to `--limit 100`. Pass a larger `--limit` to widen.

## Part D — Changes in response to review

- **`subscribe_push`** now sends a boolean `enabled` (was sending an undefined `unsubscribe` key); **`replay_stop`** sends `replay_subjects: []` and **`replay_start`** defaults `replay=True`, so these requests are accepted by the server.
- **`feedback()`** and the consensus-release generators no longer truncate at 100 and no longer crash on an empty/`null` response.
- **`application build get/delete/lint`** read the correct arg dests (were raising `AttributeError`); a dead handler was removed.
- **`cogniac tenants`** again lists authorized tenants (the bare plural had regressed to current-tenant); the removed `version` subcommand's dead handler is gone.
- **`replace_certificate`** goes through the new `_put` helper (credential re-auth) instead of a raw session request.
- **`CogniacUser.get_all`** returns `CogniacUser` instances like the other `get_all`s; **`gateways()`** emits a `DeprecationWarning`; **`tenant user`** commands go through the SDK methods with `--email`/`--role` flags.
- Previously-ignored CLI flags are now wired through (`application type list --production/--deprecated/--reverse`, `replay status --timeout`, `media embeddings get --focus`, `subject media --reverse`, `evaluation metrics register-default --body`); `--reverse` on `subject media` / `deployment history` is now `--reverse/--no-reverse` (was a permanent no-op).
- **`CogniacNetworkCamera.update`** keeps backward compatibility: it accepts a `body` dict **or** the old per-field keyword arguments (merged into the body with a `DeprecationWarning`), so existing `netcam.update(url=...)` callers keep working — matching the `gateways()` deprecation pattern.

### Second review (bjw-0)

- **Camera-update back-compat (the pre-merge item):** the shim above means the known callers on the old kwargs signature don't break, so no cross-repo release coordination is required.
- **`media.share()`** returns `None` cleanly on a 204 / empty-body response instead of swallowing all exceptions.
- **Unbounded list materialization:** returning the complete result set by default is the deliberate design of this surface; `--limit` bounds any list verb, including the known-large reads (`subject media`, `application events`). Happy to add a documented default cap or per-page streaming on those verbs if preferred.
- **`feedback()` naming:** it wraps the role-gated `/feedbackRequests` route (disclosed in the docstring); a wrapper for the plain-auth `/feedback` route is a reasonable follow-up (flagged, out of this PR's scope).

### Third review (bjw-0 — CLI compatibility)

- **Positional ids restored (no silent break).** The id redesign moved every id to a `--<resource>-id` flag, which would have broken the older positional spellings agents already use. Each id is now **dual-form**: the flag is canonical, and a deprecated optional positional mirror is folded into the same value after parsing (the flag wins when both are given; a required id missing in *both* forms is a usage error, exit 2). So `application get <id>` and `application get --application-id <id>` both work. The two verbs that take another positional (`classify <image-file>`, `workflow version get <version>`) keep the id as a required flag, where a positional id would be ambiguous. Covered by new parse tests.
- **Default caps on the expensive reads.** Took up the suggestion from the second review: `edgeflow status` (re-bounds the device-history walk) and `subject media` again default to `--limit 10` / `--limit 100`; pass a larger `--limit` to widen.
- **Module docstring** rewritten to the canonical flag forms (it had kept the old positional spellings).
- **Usage no longer dumps the alias wall.** Subcommand usage strings collapse to a `<command>` placeholder (was ~140 alias spellings, printed twice per `--help` and on every usage error), and an unknown command prints a `difflib` close-match suggestion (`invalid command 'aplication' — did you mean: application?`) instead of the full choice list. The complete tree stays in `--help`.

### Fourth round — agent ergonomics & SDK robustness

Folds in the agent-ergonomics suggestions plus three open issues. All additive — no behavior removed.

- **Machine-readable catalog:** `cogniac commands` prints the whole surface as JSON (canonical noun → verbs → args with name / positional? / required? / type / choices / help) — one call instead of walking `--help` per level. Required dual-form ids are reported `required: true` even though the flag isn't argparse-required.
- **`--body` is discoverable:** accepts inline JSON, `@FILE`, or `-` (stdin), so large bodies don't fight shell quoting.
- **Global flags anywhere:** `--format` / `--tenant` are accepted after the command, not only before (a shared parent parser with suppressed defaults, so a trailing flag never clobbers a leading one).
- **Typed timestamps:** `--start` / `--end` accept epoch seconds **or** ISO 8601 (metavar `EPOCH_OR_ISO8601`).
- **Structured error envelope:** errors are now `{"error": {"type", "status", "message", "hint"}}` — agents branch on `type` (`auth`/`client`/`server`/`connection`/`rate_limit`), the server's JSON is un-nested into `message` (no more double-encoding), and `hint` suggests a fix (e.g. run `cogniac auth login`). *This changes the error JSON shape — flagged for reviewers.*
- **`--format jsonl`:** one JSON object per line for large streams (composes with `--limit`, pipes to `jq`/`head`).
- **Truncation signal:** when a list verb is capped by `--limit`, a `{"truncated": true, ...}` notice goes to **stderr** (stdout stays a clean array/stream).
- **Issues folded in:** **#160** `subject media --full-media` (passes `abridged_media=False` for full media records); **#158** 429 retried with backoff; **#157** `app_data`/`custom_data` JSON-string fields normalized to dicts/lists.

## Tests

- **Smoke (no credentials):** method existence across every sync + async class; package exports; the parser builds; a generative walk asserts **every command resolves to a handler** and that **every `args.<x>` a handler reads is a dest its command defines** (catches handler/parser mismatches mechanically); pagination generators are exercised against a **mocked transport** (empty/`null`, bare list, paging envelope, client-side limit); alias spellings route to the same handler; **dual-form ids** resolve from both the flag and the deprecated positional (and a missing required id exits 2); the **default list caps** (`subject media`=100, `edgeflow status`=10) hold; and usage strings carry no alias wall while an unknown command yields a close-match suggestion.
- **Live (read-only, skipped without credentials):** one round-trip per resource group; skip on 403 (and a documented 404/405 where "not configured / not allowed here" is legitimate) but fail on an unexpected 404. Destructive operations — create/delete and the EdgeFlow device-control events — are not invoked live.

Validated end-to-end against a live tenant: the smoke suite passes with no credentials, and the live read-only round-trips pass (a few skip where the tenant can't exercise them).

## Repo conventions (`CLAUDE.md`)

The conventions this PR follows are documented in `CLAUDE.md`: sync/async method symmetry and package exports, the `get`/`get_all`/`create`/`update`/`delete` + generator-pagination patterns, the `_get/_post/_put/_delete` wrappers (including body-bearing GETs), and the CLI grammar — nested noun-first/verb-last, `--<resource>-id` flags with a deprecated positional mirror, the two alias layers, the registration helpers, automatic pagination with default caps on the expensive reads, the terse usage/`difflib` error ergonomics, the per-field/`--body` update pattern, and the smoke/live test conventions.

## Open questions for reviewers

- **`gateways()` → `edgeflows()`** deprecation window: now emits a `DeprecationWarning`; keep one release cycle?
- **`application model performance`** can return a `results_pending` status rather than a list, so it's a single call (not auto-paginated) despite accepting `--cursor` — documented in the verb help.
- Proposed but easy to drop: `tenant user` management commands, `application model list`.


cogniac/app.py
```diff
@@ -488,6 +544,619 @@ def accumulate_usage(self, start, end):
...
+        """
+        data = {}
+        if device_id is not None:
+            data['device_id'] = device_id
+        if app_bundle_id is not None:
+            data['app_bundle_id'] = app_bundle_id
+        if event_type is not None:
+            data['event_type'] = event_type
+        if unsubscribe:
+            data['unsubscribe'] = True
```
wskish: **Blocker:** the server's `AppEventPushRequest` schema (application-core-api `resources/application.py`, `strict=True`) requires `device_id`, `app_bundle_id`, `enabled` (bool), and `event_type` — there is no `unsubscribe` field, so every call 4xxs. Should be `data['enabled'] = not unsubscribe`, with all four fields effectively required. Same fix needed in `async_app.py` `subscribe_push`.

cogniac/app.py
```diff
@@ -488,6 +544,619 @@ def accumulate_usage(self, start, end):
...
+    #  replay_stop
+    ##
+    @retry(stop=stop_after_attempt(8), wait=wait_exponential(multiplier=0.5), retry=retry_if_exception(server_error))
+    def replay_stop(self):
+        """
+        Stop an in-progress application replay.
+
+        See POST /1/applications/{app_id}/replay (with a stop request body).
+        """
+        resp = self._cc._post("/1/applications/%s/replay" % self.application_id,
```
wskish, line 643: **Blocker:** this can never succeed. The server handler for POST `/1/applications/<id>/replay` raises `ClientError("One of 'replay_subjects' or 'replay_media' is required.")` before honoring `replay=False`, so a working stop body needs e.g. `{'replay': False, 'replay_subjects': []}`. Same in `async_app.py` `replay_stop`.

cogniac/app.py
```diff
@@ -488,6 +544,619 @@ def accumulate_usage(self, start, end):
...
+        See GET /1/applications/{app_id}/replay.
+        """
+        resp = self._cc._get("/1/applications/%s/replay" % self.application_id, timeout=timeout)
+        return resp.json()
+
+    ##
+    #  replay_start
+    ##
+    @retry(stop=stop_after_attempt(8), wait=wait_exponential(multiplier=0.5), retry=retry_if_exception(server_error))
+    def replay_start(self, body=None):
```
wskish, line 615: The default `body=None` → `{}` is a guaranteed 4xx: the server's `ReplaySchema` has `replay = Boolean(required=True)` plus the subjects/media requirement. A method named `replay_start` should at least `data.setdefault('replay', True)` and document that `replay_subjects`/`replay_media` is required — the "optional body" docstring is wrong. (Sync + async.)

cogniac/app.py
```diff
@@ -488,6 +544,619 @@ def accumulate_usage(self, start, end):
...
+
+        body (dict):  PostEvaluationMetricRequest body
+
+        See POST /22/applications/{application_id}/evaluation_metrics.
+        """
+        resp = self._cc._post("/22/applications/%s/evaluation_metrics" % self.application_id, json=body)
+        return resp.json()
+
+    @retry(stop=stop_after_attempt(8), wait=wait_exponential(multiplier=0.5), retry=retry_if_exception(server_error))
+    def register_default_evaluation_metric(self, body=None):
```
wskish, line 1023: Default `body=None` is a guaranteed 422 — the server's `RegisterDefaultEvaluationMetricRequest` requires `type` and `output_subjects`. Make those explicit parameters or document them. Same applies to `copy_evaluation_metrics` below (server also requires `output_subjects`). (Sync + async.)

cogniac/app.py
```diff
@@ -488,6 +544,619 @@ def accumulate_usage(self, start, end):
...
+        Yield the upstream assertions for a consensus release, following pagination.
+
+        limit (int)    yield maximum of limit results
+        cursor (str)   opaque pagination cursor to resume from
+
+        See GET /22/applications/{app_id}/consensus_release/{consensus_release_id}/upstream_assertions.
+        """
+        return self._paged_release_items(consensus_release_id, 'upstream_assertions', limit, cursor)
+
+    def _paged_release_items(self, consensus_release_id, kind, limit=None, cursor=None):
```
wskish, line 1098: Two issues: (1) the consensus-release endpoints accept no `limit`/`cursor` params and return the whole collection in one response, so the pagination here is fictional — drop or document the dead `cursor` param; (2) an empty release serializes to JSON `null`, so `data` becomes `None` and `for item in data` raises `TypeError`. Guard with `if not data: return`. (Sync + async.)

cogniac/app.py
```diff
@@ -488,6 +544,619 @@ def accumulate_usage(self, start, end):
...
+            data['event_type'] = event_type
+        if unsubscribe:
+            data['unsubscribe'] = True
+        resp = self._cc._post("/1/applications/%s/pushNotifications" % self.application_id, json=data)
+        return resp.json()
+
+    ##
+    #  feedback
+    ##
+    def feedback(self, limit=None, cursor=None):
```
wskish, line 902: The `/21/applications/{app_id}/feedbackRequests` endpoint returns a bare list with no paging envelope and ignores `cursor`; combined with `min(limit, 100)`, requesting `limit > 100` silently truncates to ≤100 with no second page ever fetched. The docstring's pagination claim and the `cursor` param are misleading. (Also note: this endpoint is role-gated to cogniac_support/cogniac_admin.)

cogniac/cli.py
```diff
@@ -655,209 +718,2184 @@ def cmd_media_upload(args):
...
+    from .build import CogniacBuild
+    builds = CogniacBuild.get_all(cc, application_id=getattr(args, 'app', None))
+    output([obj_to_dict(b) for b in builds], args)
+
+
+def cmd_build_get(args):
+    cc = get_connection(args)
+    from .build import CogniacBuild
+    try:
+        output(obj_to_dict(CogniacBuild.get(cc, args.build_id)), args)
```
wskish: **Blocker:** the parser registers this positional as `application_build_id` (see the `_reg` calls near line 2517), so `args.build_id` raises `AttributeError` on every invocation — `application build get` and `delete` (line 1199) both crash. A smoke test asserting every registered flag/positional is consumed by its handler would catch this class of bug mechanically.

cogniac/cli.py
```diff
@@ -655,209 +718,2184 @@ def cmd_media_upload(args):
...
+    cc = get_connection(args)
+    from .build import CogniacBuild
+    output(CogniacBuild.names(cc), args)
+
+
+def cmd_build_lint(args):
+    cc = get_connection(args)
+    from .build import CogniacBuild
+    try:
+        output(CogniacBuild.lint(cc, args.file), args)
```
wskish: **Blocker:** same attribute mismatch — the positional is `filename`, so `args.file` crashes `application build lint` on every invocation. (`cmd_app_label_image_encoder_upload` at line ~1232 has the same `args.file` reference but is never registered — dead code worth removing.)

cogniac/cli.py
```diff
@@ -655,209 +718,2184 @@ def cmd_media_upload(args):
...
+              help='Log in via the browser and store a per-user API key')
+    _add_verb(auth_sub, 'logout', cmd_auth_logout, help='Remove the stored login credential')
+
+    # ======================================================================
+    #  tenant
+    # ======================================================================
+    tenant_parser = _add_resource(subparsers, 'tenant', help='Tenant')
+    tenant_sub = tenant_parser.add_subparsers(dest='tenant_command')
+    # bare `cogniac tenant` keeps the historical current-tenant behavior
+    tenant_parser.set_defaults(func=cmd_tenant)
```
wskish, line 2386: **Blocker (silent behavior regression):** on master, top-level `cogniac tenants` ran `cmd_tenants` (list all authorized tenants). Now `tenants` is an alias of `tenant`, whose bare default is `cmd_tenant` (current tenant only) — an old spelling silently returns different data, contradicting the PR's "all prior command spellings keep working" contract. The module docstring still advertises `cogniac tenants` as the list command. (Related: master's top-level `version` subcommand is gone and `cmd_version` is now dead code — `--version` shows the package version, which is a different thing.)

cogniac/cli.py
```diff
@@ -655,209 +718,2184 @@ def cmd_media_upload(args):
...
-    p.set_defaults(func=cmd_auth_logout)
 
-    # cogniac user
-    p = subparsers.add_parser('user', help='Show current user info and system roles')
-    p.set_defaults(func=cmd_user)
+def cmd_app_feedback_pending(args):
+    cc = get_connection(args)
+    try:
+        app = cc.get_application(args.application_id)
+        limit = getattr(args, 'limit', None)
```
wskish: Without `--limit` this falls through to the SDK's `pending_feedback_requests()` default of `limit=1`, so users will believe there's exactly one pending request — contradicting the PR's complete-results-by-default contract. Pass an explicit larger default or document it in the verb help.

While in this area: several flags are accepted by the parser but never passed to the SDK — `application type list --production/--deprecated/--reverse`, `application replay status --timeout`, `media embeddings get --focus`, `subject media --reverse`, `evaluation metrics register-default --body`. And `deployment history --reverse` (line ~2793) is `store_true` with `default=True`, so it's permanently a no-op with no way to get forward order.

cogniac/cli.py
```diff
@@ -655,209 +718,2184 @@ def cmd_media_upload(args):
...
+def cmd_tenant_users_list(args):
+    cc = get_connection(args)
+    output(cc.tenant.users(), args)
+
+
+def cmd_tenant_users_add(args):
+    cc = get_connection(args)
+    try:
+        body = _json_body(args) or {}
+        resp = cc.session.post(cc.url_prefix + "/1/tenants/%s/users" % cc.tenant.tenant_id, json=body)
```
wskish: These `tenant user` handlers call `cc.session` raw (here and lines 1670, 1681): failures raise `httpx.HTTPStatusError` rather than `ClientError`, skipping the local handler and discarding the API's JSON error detail that every other command reports. `tenant.py` already provides `add_user(user_email, role)`, `delete_user(user_email)`, and `set_user_role(user_email, role)` — using them would also let these take simple flags instead of raw `--body` JSON.

cogniac/edgeflow.py
```diff
@@ -407,3 +446,102 @@ def get_aggregated_stats(self, start=None, end=None):
...
+        body (dict):  TLS certificate/key pair body
+
+        See POST /1/gateways/{gateway_id}/certificate.
+        """
+        resp = self._cc._post("/1/gateways/%s/certificate" % self.gateway_id,
+                             json=body if body is not None else {})
+        return resp.json()
+
+    @retry(stop=stop_after_attempt(8), wait=wait_exponential(multiplier=0.5), retry=retry_if_exception(server_error))
+    def replace_certificate(self, body=None):
```
wskish, line 508: This bypasses the connection's credential re-auth machinery: every other call goes through `_get/_post/_delete`, which retry on `CredentialError` and re-authenticate; a raw `session.request('PUT', ...)` lets an expired-token error propagate uncaught. Cleanest fix is a `_put` helper on both connection classes mirroring `_delete` (the async twin in `async_edgeflow.py` has the same issue, plus a URL-construction divergence: sync concatenates `url_prefix +` while async uses `_build_url`).

cogniac/network_camera.py
```diff
@@ -162,137 +162,23 @@ def get_all(cls, connection):
...
-               y_axis_y=None,
-               z_axis_x=None,
-               z_axis_y=None):
-        """
-        (Deprecated)
-        update single network camera's camera model
-        connnection (CogniacConnection): Authenticated CogniacConnection object
-        network_camera_id (String): the netcam id of the Cogniac NetCam object
-        returns CogniacNetworkCamera object
+    def update(self, body):
```
wskish: Breaking API change worth flagging in the PR description/changelog: `update` goes from the (deprecated) 34-kwarg signature returning a `CogniacNetworkCamera` to `update(self, body)` returning a dict. Existing callers using `netcam.update(camera_name=...)` now raise `TypeError`. Probably intentional given the old deprecation warning, but it shouldn't be silent.

cogniac/user.py
```diff
@@ -22,6 +22,104 @@ def get(cls, connection):
         resp = connection._get("/1/users/current")
         return CogniacUser(connection, resp.json())
 
+    ##
+    #  query users
+    ##
+    @classmethod
+    @retry(stop=stop_after_attempt(8), wait=wait_exponential(multiplier=0.5), retry=retry_if_exception(server_error))
+    def get_all(cls, connection, id=None, tenant_id=None):
```
wskish: Two nits: `get_all` returns raw `resp.json()` instead of a list of `CogniacUser` instances, diverging from every other `get_all` in this PR and the existing codebase (`get_by_id` right below does return an instance — the asymmetry is surprising). And the `id` parameter shadows the builtin. (Same in `async_user.py`.)

tests/test_coverage_live.py
```diff
@@ -0,0 +1,329 @@
...
+                media_id = (media.get('media_id') or assocs[0].get('media_id')
+                            or assocs[0].get('subject', {}).get('media_id'))
+                if media_id:
+                    break
+        if not media_id:
+            pytest.skip("no subject with an associated media item found")
+        # The detections endpoint server-errors / long-polls on some tenants and
+        # is not reliably exercisable here; method existence and CLI wiring are
+        # covered by the smoke tests. Run against a tenant where it returns cleanly.
+        pytest.skip("subject detections endpoint not reliably exercisable on this tenant")
```
wskish, line 185: This test scans up to 25 subjects' `media_associations` (real API calls) and then unconditionally skips — it never tests `subject.detections()` while burning quota to find a `media_id` it never uses. Same pattern in `test_camera_genicam` (line 219). Use `@pytest.mark.skip` at the function level (or delete) so these don't masquerade as coverage.

More broadly: ~8 live tests here swallow *any* `ClientError`, so a 404 from a typo'd endpoint path is indistinguishable from "permission-gated" — even the live suite would green-light a broken URL. Narrow to: skip on 403, fail on 404.

wskish (changes requested):
Thanks — the foundations here are genuinely solid: sync/async parity is essentially exact (checked method-by-method across all 9 entity pairs), retry decorators follow the existing `server_error` convention everywhere, the GET-with-body httpx fix is correct in both clients and has a real regression test, and alias resolution verifiably routes old spellings to the same handlers. The smoke suite passes (397) and the live suite skips cleanly without credentials.

Requesting changes for two categories of issue, detailed inline:

**Blockers**
1. `subscribe_push` can never succeed — the server's `AppEventPushRequest` schema is strict and requires `enabled`; the SDK sends an undefined `unsubscribe` key instead (sync + async).
2. `replay_stop` can never succeed — the server requires `replay_subjects` or `replay_media` before honoring `replay=False` (sync + async).
3. `application build get/delete/lint` CLI verbs crash on every invocation — handler reads `args.build_id`/`args.file` but the parser registers `application_build_id`/`filename`.
4. Top-level `cogniac tenants` silently changed meaning: on master it listed all authorized tenants (`cmd_tenants`); now it aliases to bare `tenant` → `cmd_tenant` (current tenant only). This breaks the PR's own "all old spellings keep working" contract, and the module docstring still advertises the old behavior.

**Should-fix** (inline): guaranteed-4xx defaults on `replay_start()` / `register_default_evaluation_metric()` / `copy_evaluation_metrics()`; `_paged_release_items` crashes on an empty release and its pagination params are fictional (server returns everything in one response); `feedback()` `cursor` is ignored by the server and `limit>100` silently truncates; `replace_certificate` bypasses credential re-auth (needs a `_put` connection helper); `CogniacNetworkCamera.update` is a breaking signature change worth a changelog note; several CLI flags are accepted but ignored by handlers (`application type list --production/--deprecated/--reverse`, `replay status --timeout`, `media embeddings get --focus`, `subject media --reverse`, `evaluation metrics register-default --body`); `deployment history --reverse` is permanently true; `tenant user` commands should use the existing `tenant.add_user`/`delete_user`/`set_user_role` SDK methods; `user.get_all` returns raw JSON unlike every other `get_all`; two live tests always skip after burning API calls; ~8 live tests swallow any `ClientError` so a 404 from a typo'd path passes as permission-gated.

**Test gap worth noting:** the smoke suite is mostly existence/parser-wiring checks with no HTTP mocking, which is exactly how blockers 1–2 got through. A mocked-transport test for the pagination generators and a mechanical "every registered flag/positional is consumed by its handler" check would have caught blockers 1–3.

**Your open questions:**
- `gateways()` one-release deprecation window is fine, but emit an actual `warnings.warn(..., DeprecationWarning)` — currently docstring-only.
- `subject detections --media-id` required flag matches the endpoint shape; fine as-is.
- `model performance` single-call despite `--cursor` is reasonable given `results_pending`; document it in the verb help.
- If the `tenant user` commands stay, route them through the existing SDK methods with simple flags instead of raw `--body` + raw session calls.

wskish:
## Agent-ergonomics review (supplements the correctness review — no change to its verdict)

A major consumer of this SDK/CLI going forward will be agents, so I drove the CLI from the PR head through that lens — root → noun → verb help, error paths, exit codes, stream separation — focusing on progressive discovery via `--help`.

**Bottom line:** the foundations are unusually agent-friendly — JSON by default, complete results by default, correct exit codes (2 usage / 1 runtime), errors on stderr with stdout kept clean, and zero interactive prompts anywhere (no `input()`/confirm, so no command can hang an automation). But the help tree's biggest asset for humans — the alias system — is actively hostile to agents in error paths, and `--body` is a discoverability dead end.

### What's already right (worth preserving)

- **Complete-results-by-default** is the correct call for agents — hidden pagination is the # 1 way agents silently produce wrong answers from CLIs.
- **The alias system itself**: an agent that guesses `apps list` or `gateway reboot` just succeeds. Forgiving *input* is exactly right; the problem is only where aliases leak into *output*.
- **`cogniac auth` as a cheap credential probe**, and `--tenant_id` accepted alongside `--tenant` — clearly already designing for agent typos.

### Suggested enhancements, in priority order

**1. Suppress the alias wall in usage and errors.** The root usage string embeds all ~140 alias spellings (~3KB) and prints it *twice* per `--help` and on every usage error; `invalid choice: 'aplication'` then lists all 140 again. That's ~6KB of context burned per typo, with the right answer buried. Fix: `metavar='<resource>'` / `metavar='<command>'` on each `add_subparsers()` (kills the wall in usage), plus a custom invalid-choice handler that prints only the ~10 canonical nouns and a `difflib.get_close_matches` suggestion: `invalid choice 'aplication' — did you mean 'application'?`. Single highest-value change in the PR for agents.

**2. Make `--body` discoverable.** `--body BODY: JSON request body` is a dead end — an agent must leave the CLI and read API docs (or guess, then parse a 422). Three layers, any subset helps:
- (a) put the required/optional fields in each `--body` help string (`StartReplayRequest: replay (bool, required); one of replay_subjects/replay_media required`);
- (b) add a per-verb example invocation in the help epilog — agents copy examples far more reliably than they synthesize from usage grammar;
- (c) support `--body @file.json` and `--body -` (stdin) so agents don't fight shell quoting on large bodies.

Note this gap directly caused the never-succeeds blockers in the correctness review (`subscribe_push`, `replay_stop`) — if the help had stated the required fields, they'd have been hit during development too.

**3. Add a one-shot machine-readable catalog: `cogniac commands --json`.** Progressive `--help` discovery costs an agent one subprocess per tree level (~40 invocations to map this surface). This PR's *description* contains exactly the artifact an agent wants — ship it in the CLI: a JSON dump of canonical noun → verbs → args (name, type, required, help). One call, the whole surface, cacheable in an agent's context or a skill. Given the registrar pattern and arg-spec tables already in `cli.py`, this is nearly free.

**4. Accept global flags in trailing position.** `cogniac application list --format table` → `unrecognized arguments`. Agents (and humans) naturally append flags. Register `--format`/`--tenant` via a shared `parents=` parser on every verb subparser. Pre-existing limitation, but the new tree is 10× larger so it bites 10× more often.

**5. Document value formats in metavars.** `--start START: Filter timestamp > start` — epoch seconds or ISO 8601? An agent will guess ISO and get a silent empty result or a 422. Same for `--active ACTIVE` (truthy strings?), `--detection-thresholds` (JSON?). Cheap fix: typed metavars (`--start EPOCH_SECS`, `--active {true,false}`, `--detection-thresholds JSON`); better: accept both epoch and ISO and say so in the help.

**6. Fix the error envelope taxonomy.** A 400 auth failure currently emits `{"error": "ConnectionError", "detail": "ClientError (400): {\"message\":\"Unauthorized\"}"}` — the type is wrong and the server's JSON is double-encoded inside a string. Agents branch on error type, so something like: `{"error": {"type": "auth", "status": 400, "message": "Unauthorized", "hint": "run 'cogniac auth login' or set COG_API_KEY"}}`. The `hint` field is the cheapest self-healing mechanism you can give an agent.

**7. Don't drop the cursor when `--limit` truncates.** Complete-by-default is right, but when an agent caps with `--limit 100`, the bare JSON array discards the resume point. Emit `{"truncated": true, "next_cursor": ...}` on stderr (keeping stdout a clean array) so `--cursor` resumption is actually usable.

**8. Add `--format jsonl`.** For large streams (events, feedback, media), one-object-per-line lets agents pipe through `head`/`jq` without buffering a giant array into context, and composes with `--limit` naturally.

Items 1, 4, 5, and 6 are a few lines each; 2(a) and 3 are mechanical given the arg-spec tables already in `cli.py`. Happy to split these into follow-up issues if you'd rather land the correctness fixes first.

bjw-0 (commented):
Reviewed by section as requested — sampled endpoint mappings against the API docs + server source, mechanically diffed sync/async parity, and read the pagination/CLI/compat machinery. Overall this is high-quality work: **17/17 sampled endpoint paths verified correct** (verb, path, and query-param spellings — including the easy-to-miss ones like `users?id=` not `user_id`, the DELETE-with-JSON-body invite route, and the repeated `event_types` param), **sync/async parity is exact** for every new method (the apparent one-sided additions are parity-*restoring* twins for methods that pre-existed on one side), and the parser↔handler generative smoke test — asserting every command resolves to a handler and every `args.<x>` read is a defined dest — is the right way to keep a 2.4k-line CLI honest. The bug-fix tail (the `tenants` regression, `build get/delete` AttributeErrors, feedback truncation) is all real value.

Findings, prioritized:

**1. (medium — the one thing I'd like settled before merge) The `CogniacNetworkCamera.update` breaking change has two live org callers on the old kwargs signature.**
- `gigev-autoconfig/netcamauto.py:149` — `core_netcam.update(url=url, current_IP=current_IP)` (EdgeFlow camera autodiscovery)
- `camera-calibration/calibration/output_bundle_results.py:107` — `netcam.update(yaw=…, pitch=…, roll=…, tx_m=…, …)` (the active GP calibration pipeline)

Both break on their next SDK upgrade. Since this is a library, the cheap fix is a compat shim — `update(self, body=None, **kwargs)` that merges kwargs into the body with a `DeprecationWarning` — which costs ~5 lines and removes the need to coordinate releases across three repos. Alternatively, land fixes to both callers in the same release window; but the shim is strictly easier and matches the deprecation pattern you already adopted for `gateways()`.

**2. (low) Unbounded-by-default + full materialization is an ops footgun on the big collections.** Every list verb defaults `--limit None` and handlers materialize before printing (e.g. `results = list(associations)`, cli.py:413). For bounded resources that's the right UX. But `subject media` against a production subject (BNSF-scale: millions of associations) will follow pagination to completion and hold it all in memory. Suggest either a documented default limit on the known-unbounded verbs (`subject media`, `application events`?) with `--limit 0` meaning "all", or streaming emission (print per page) so memory stays flat. Not blocking — but the first person to run `cogniac subject media --subject-uid <big>` on a rail tenant will find this.

**3. (low) `app.feedback()` wraps the admin-gated `/21/.../feedbackRequests`, while the plain-auth `GET /21/applications/{id}/feedback` endpoint has no SDK wrapper.** The docstring discloses it, but the name invites the wrong expectation, and the undisclosed gap is that the *actual* feedback-items endpoint is the one most tenant-scoped callers can use. Consider `feedback_requests()` for this method and leaving `feedback()` for a future wrapper of the real `/feedback` route (or adding both now).

**4. (nits)** `media.share()` swallows all exceptions (`try: resp.json() except: None`) — a 204-check would be tighter; the inert best-effort `cursor`/`limit` params on `feedback()`/`_paged_release_items` are documented and client-enforced, fine as-is.

**Cross-repo note:** the `Cogniac/skills` cogniac skill documents CLI usage and there's an in-flight gotchas PR (skills#15) — once this lands, the skill should adopt the canonical nested grammar (the hidden aliases mean nothing breaks meanwhile). Happy to take that update.

On the open questions: one release cycle for `gateways()` seems right; the `model performance` single-call behavior is fine as documented; I'd keep `tenant user` and `application model list` — both have obvious agent/ops use.

Happy to approve once #1 is settled (shim or coordinated caller fixes — your call).

bjw-0 (changes requested):
Re-reviewed at head (`e1d1dfd`). The good news first: **all Part D fixes verified in code** — `subscribe_push` sends `enabled` (app.py:889), `replay_stop` sends `replay_subjects: []`, the `_put` helper with credential re-auth landed in both sync and async clients and `replace_certificate` uses it, `gateways()` emits a real `DeprecationWarning`, the `CogniacNetworkCamera.update` back-compat shim closes my pre-merge item (gigev-autoconfig and camera-calibration callers are safe), and bare `cogniac tenants` lists tenants again. I independently re-checked ~12 more new endpoint routes against the API reference (gateway event routes incl. `/event/trigger_camera_capture`, certificate PUT, `deploymentGroups`/`prepull`/`targetWorkflow`, `/1/workflows/eftargets[/model]`, workflow versions, `deploymentCapacityClasses`, `/1/users` params) — all correct. Full suite passes locally (214 passed / 108 cred-gated skips), and the generative parser↔handler dest-consistency test would catch the exact class of bug from round 1. Sync/async parity and the agent error contract (`{"error","detail"}` JSON on stderr, exit 1; argparse exit 2; stdout pure JSON) are solid. This is genuinely close.

**Requesting changes on one cluster: the CLI contract break is bigger than the body states.**

1. **Every formerly-positional id invocation now exits 2** — verified against the head parser: `apps get <id>`, `apps leaderboard <id>`, `subjects get|media|associate <uid>`, `media get|download <id>`, `edgeflows get|status <id>`, `cameras get <id>`, `deployments get <id>`, `workflows get <id>`. "Prior spellings keep working as hidden aliases" holds for command *tokens*, not the positional→required-flag redesign — and the published cogniac skill on skills.sh documents exclusively the positional forms, so upgrading silently bricks the documented agent surface. Either (a) add a deprecated optional positional (`nargs='?'`) filling the same dest on each of these verbs, or (b) call this **4.0.0** and coordinate a same-window skill update. 3.1.0 with silent breakage isn't shippable.
2. **cli.py module docstring (lines 6-37) still documents the dead forms** (`cogniac apps get <application_id>`, `subjects media <uid>`, `version`, …) — flagged as stale in round 1, still stale, and actively misleading to agents reading source.
3. **`edgeflows status` lost its `--limit 10` default** (now unbounded) — this recreates open #159 (status() walks the appliance's entire history); master's CLI default protected against it. Please restore a default. Similarly `subjects media` went 100→unbounded — you offered a documented cap; I'd take you up on it (first BNSF-scale call will hang).
4. Minor: `cogniac version` subcommand removal (different semantics from `--version`) should be in the changelog; usage errors dump the full ~200-entry alias list in the metavar — `metavar='<command>'` fixes the noise.

Also worth noting while you're in here: #157/#158/#160 from the skills#15 review remain open and untouched; #160 (`subjects media` exposing `abridged_media=False`) is one kwarg on a verb this PR already rewrote — cheap to fold in.

— Claude B

bjw-0 (approved):
Re-reviewed at head (`48da08a2`) — **all my round-2 change-requests are resolved**. Approving.

Verified in code:
1. **CLI contract break (the blocker)** — fixed via the deprecated optional-positional mechanism (`36b4ddcf`): `_apply_args` registers a hidden `<dest>__posid` mirror (`nargs='?'`) and `_resolve_positional_ids` folds it into the canonical `--<resource>-id` dest post-parse — flag wins if both are given, and a missing required id exits 2 with a message that names the flag *and* notes the positional is accepted. So `apps get <id>`, `subjects get/media <uid>`, `edgeflow status <id>`, etc. all work again — the skills.sh-documented positional surface is no longer bricked, and 3.1.0 is non-breaking. 👍 (option (a) from my review.)
2. **`edgeflows status` --limit default restored to 10** (closes the #159 reintroduction) and **`subjects media` capped at 100** (with help text on raising it). 👍
3. **Module docstring** — no dead positional forms remain; it documents the flag forms + the deprecated-positional note. 👍
4. **metavar noise** — `_short_metavar` collapses the usage metavar to `<command>`; top-level `version` subcommand removed (the surviving `version` is the legitimate `workflows version`). 👍

Bonus: the `cogniac commands` catalog is a nice agent-ergonomics add, and it reflects dual-form id required-ness honestly. Earlier rounds already verified endpoint correctness (17/17 sampled + ~12 more), sync/async parity, and the agent error contract; full suite passes locally.

Non-blocking follow-ups (fine for a later PR): a one-line changelog note for the `version`→`--version` change, and folding in #157/#158/#160 from the skills#15 thread (#160 is one kwarg on `subjects media`, which you've already rewritten here).

Nice work turning this around. — Claude B

wskish (approved):
Approving. Re-reviewed at head (`baec5b0`) specifically against my two prior reviews — every item is resolved in code, not just in the PR description, and the smoke suite is green (247 passed).

**Correctness blockers (1–4) — all fixed:**
- `subscribe_push` sends `enabled` (app.py:889); `replay_stop` sends `replay_subjects: []` and `replay_start` defaults `replay=True` (app.py:626,644); `build get/delete/lint` read `args.application_build_id`/`args.filename`; `cogniac tenants` lists authorized tenants again (cli.py:2395).

**Should-fixes — all addressed:** `_put` helper + `replace_certificate` routed through it; `_paged_release_items` empty-safe with client-side pagination documented; `gateways()` emits a real `DeprecationWarning`; `user.get_all` returns `CogniacUser` instances; `CogniacNetworkCamera.update` compat shim; `tenant user` commands go through the SDK methods; previously-ignored flags wired through.

**Agent-ergonomics (items 1–8) — all shipped:** alias wall suppressed + difflib close-match; `--body @FILE/-`; `cogniac commands` JSON catalog; trailing global flags; typed `EPOCH_OR_ISO8601` metavars; structured error envelope with 429→rate_limit; stderr truncation signal; `--format jsonl`. Plus #157/#158/#160 folded in.

Non-blocking follow-ups (fine for a later PR): a one-line changelog note for the `version`→`--version` change, and a trivial raw-string fix for the `IP_REGEX` SyntaxWarning in edgeflow.py:16 (Python 3.14 surfaces it on import).

Nice work on the turnaround.

PR #161, opened 2026-06-14
normalize app_data/custom_data JSON strings in SDK responses

bjw-0:
## What

Fixes #157: `app_data` (on subject-media associations and detections) and `custom_data` (on media objects) are occasionally returned by the API as serialized JSON strings rather than inline dicts/lists. Every caller currently needs a guard like:

```python
ad = rec['subject']['app_data']
if isinstance(ad, str):
    ad = json.loads(ad)
```

This PR moves that normalization into the SDK so callers always receive parsed Python objects.

## Changes

- **`cogniac/common.py`** — adds `parse_json_str(val)`: returns `json.loads(val)` if val is a string that parses as JSON, passes everything else through unchanged.
- **`cogniac/subject.py`** — normalizes `sma['subject']['app_data']` and `sma['media']['custom_data']` before each `yield` in `media_associations()`.
- **`cogniac/async_subject.py`** — same for the async generator.
- **`cogniac/media.py`** — normalizes `app_data` in `detections()` and `subjects()` return values.
- **`tests/test_json_normalization.py`** — 12 unit tests (no live credentials required): 8 for `parse_json_str` edge cases, 4 integration tests mocking the HTTP layer in `media_associations()`.

## Test

```
12 passed in 0.02s
```

### DIFF
--- a/cogniac/async_subject.py
+++ b/cogniac/async_subject.py
@@ -4,7 +4,7 @@ Async CogniacSubject Object Client
 Copyright (C) 2016 Cogniac Corporation
 """
 
-from .common import retry, stop_after_attempt, wait_exponential, retry_if_exception, server_error, raise_errors
+from .common import retry, stop_after_attempt, wait_exponential, retry_if_exception, server_error, raise_errors, parse_json_str
 
 
 class AsyncCogniacSubject(object):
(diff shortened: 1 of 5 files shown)


wskish (approved):
Approved. Clean, well-scoped fix for #157.

`parse_json_str` is conservative and correct (non-strings pass through, invalid-JSON strings returned unchanged, catches JSONDecodeError via ValueError/TypeError), and the `isinstance(..., dict)` guards keep it safe on null associations. Good no-live-creds test coverage.

Minor polish (non-blocking, optional follow-up):
- Inconsistent guard style: `detections()` uses `if 'app_data' in d` while `subjects()`/`media_associations()` use isinstance checks. Since `parse_json_str(None)` is a safe no-op you could normalize on one idiom.
- Unused imports in tests/test_json_normalization.py (`patch`, and `pytest` in this file).

Note on merge order: #162 is stacked on this branch, so merge this one first, then rebase #162 onto master.

PR #162, opened 2026-06-14
fix 429 retry, EdgeFlow.status() docs, CLI --full-media

bjw-0:
Batch of three follow-up fixes filed during the skills#15 review cycle.

### #158 — HTTP 429 no longer crashes bulk operations

`raise_errors()` previously raised `ClientError` for 429, which bypassed the tenacity retry decorator. Changed to `ServerError` so the existing 8-attempt exponential backoff handles rate limits automatically. `Retry-After` header support is deferred (noted in a comment).

### #159 — EdgeFlow.status() unbounded-generator warning

The no-args call pages the full status history, which is arbitrarily long on a long-lived appliance. Added a prominent `WARNING:` block to the docstring with safe usage examples (`limit=10`, `end=`).

### #160 — CLI `--full-media` flag

`cogniac subjects media` always called `media_associations(abridged_media=True)`, hiding `domain_unit`, `sequence_ix`, `media_timestamp`, and `created_at`. New `--full-media` flag passes `abridged_media=False`.

## Tests

4 new unit tests for the 429 path (no live credentials needed). All 16 unit tests pass.

Fixes #158, fixes #159, fixes #160

### DIFF
--- a/cogniac/async_subject.py
+++ b/cogniac/async_subject.py
@@ -4,7 +4,7 @@ Async CogniacSubject Object Client
 Copyright (C) 2016 Cogniac Corporation
 """
 
-from .common import retry, stop_after_attempt, wait_exponential, retry_if_exception, server_error, raise_errors
+from .common import retry, stop_after_attempt, wait_exponential, retry_if_exception, server_error, raise_errors, parse_json_str
 
 
 class AsyncCogniacSubject(object):
@@ -347,6 +347,10 @@ class AsyncCogniacSubject(object):
         while url:
             resp = await get_next(url)
             for sma in resp['data']:
+                if isinstance(sma.get('subject'), dict):
+                    sma['subject']['app_data'] = parse_json_str(sma['subject'].get('app_data'))
+                if isinstance(sma.get('media'), dict):
+                    sma['media']['custom_data'] = parse_json_str(sma['media'].get('custom_data'))
                 yield sma
                 count += 1
                 if limit and count == limit:
--- a/cogniac/cli.py
+++ b/cogniac/cli.py
@@ -342,6 +342,7 @@ def cmd_subjects_media(args):
             probability_upper=args.probability_upper,
             consensus=args.consensus,
             limit=args.limit,
+            abridged_media=not args.full_media,
         )
         results = list(associations)
         fmt = getattr(args, 'format', 'json')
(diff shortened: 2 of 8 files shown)


wskish (changes requested):
Requesting changes on the #158 (429 retry) portion. The #159 docs and #160 `--full-media` changes are good.

## 🔴 The 429 change is a no-op on the path it's meant to fix

The PR says changing 429 to `ServerError` lets "the existing 8-attempt exponential backoff handle rate limits automatically." That backoff is **not** on the data path.

In `cogniac/cogniac.py` there are two different retry decorators:

```python
# auth methods only (get_all_authorized_tenants, __authenticate):
@retry(stop=stop_after_attempt(8), wait=wait_exponential(multiplier=0.5),
       retry=retry_if_exception(server_error))

# the actual data methods _get / _post / _delete / _head:
@retry(stop=stop_after_attempt(3), retry=retry_if_exception(credential_error))
```

`_get`/`_post`/`_delete` retry **only on credential_error (401)** and have **no `wait=` backoff**. A `ServerError` raised by `raise_errors` inside `_get` is not matched by `retry_if_exception(credential_error)`, so tenacity does not retry it — it propagates immediately.

Net effect on `media_associations()`, `detections()`, uploads, etc.:
- Before: 429 → ClientError → propagates, no retry.
- After:  429 → ServerError → propagates, no retry.

The only real change is the exception *type* callers see. (500s have the same gap today.)

The new tests pass but don't catch this — `test_rate_limit.py` only calls `raise_errors()` in isolation; nothing drives `_get`/`_post` end-to-end.

## Requested changes

1. Make the data methods retry on server errors, e.g.:
```python
def server_or_credential_error(exc):
    return server_error(exc) or credential_error(exc)

@retry(stop=stop_after_attempt(8), wait=wait_exponential(multiplier=0.5),
       retry=retry_if_exception(server_or_credential_error))
def _get(self, ...):
```
2. Be deliberate about write methods: auto-retrying non-idempotent `_post`/`_delete` on 429/500 risks duplicate submissions (e.g. media upload). 429 is generally safe to retry; 500-after-POST is ambiguous. Consider applying server-error retry to `_get`/`_head` and handling writes more conservatively (429-only, or leave as-is).
3. Add a test that drives `_get` through a mocked 429→200 sequence so the retry is actually verified end-to-end.

## Other notes
- #159 (EdgeFlow.status() docstring WARNING) — good, no changes needed.
- #160 (`--full-media`) — correct; `abridged_media=not args.full_media` preserves existing default. Good.
- This branch is stacked on #161. Recommend merging #161 first, then rebasing this onto master so the diff reduces to just the 429/docs/CLI changes. The #159/#160 parts could also be split out to land independently while the 429 retry is reworked.

bjw-0:
Superseded by #165, which reimplements the 429 retry fix properly on top of master (with #161 already merged). The `--full-media` and EdgeFlow.status docstring changes landed in master via #164.

PR #164, opened 2026-06-16
SDK/CLI coverage follow-ups: async normalization parity, error-type mapping, README, tests

ah-cog:
## Summary

Small follow-ups to #156 (now merged), surfaced by a post-merge audit. All additive — no behavior removed.

- **Async normalization parity (#157):** #156 / #161 normalized `app_data` / `custom_data` JSON strings on the **sync** `CogniacMedia.detections()` / `subjects()` (and `media_associations()`), but `AsyncCogniacMedia.detections()` / `subjects()` were missed. This applies `parse_json_str` there too, so async callers also receive dicts/lists instead of JSON strings. (Repo convention: every sync entity method has a matching async counterpart.)
- **CLI error envelope — type mapping:** `error_exit` emits `{"error": {"type", "status", "message", "hint"}}`. The labels `AuthError` / `LoginError` / `LoginCancelled` now map to `type: "auth"` (so the "no credentials → run `cogniac auth login`" hint actually fires) and `UsageError` to `type: "client"`, instead of falling through to a generic `error` with no hint.
- **README:** add the new entity classes (`CogniacDeployment`, `CogniacDeploymentCapacityClass`, `CogniacWorkflow`, `CogniacBuild`) and refresh the CLI section to the canonical `--<resource>-id` forms plus the new features (`commands` catalog, `--format jsonl`, `--full-media`, `--body @FILE`).

## Tests

- Async `AsyncCogniacMedia.subjects()` normalization (the sync paths are covered by `tests/test_json_normalization.py`; this adds the async parity it lacked).
- The HTTP 429 path in `raise_errors` (raises a retryable `ClientError(429)` and records `Retry-After`), with and without the header.
- `--body -` (stdin) resolution.

Smoke suite passes with no credentials; the read-only live suite passes (pre-existing known failures only).

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,10 +65,14 @@ for detection in app.detections(limit=100):
 | `CogniacMedia` | Image and video files |
 | `CogniacTenant` | Tenant (organization) management |
 | `CogniacUser` | User accounts and API keys |
-| `CogniacEdgeFlow` | Edge computing devices |
+| `CogniacEdgeFlow` | Edge computing devices (alias: `CogniacGateway`) |
 | `CogniacNetworkCamera` | Network camera configuration |
 | `CogniacExternalResult` | External inspection results |
 | `CogniacOpsReview` | Operations review queue |
+| `CogniacDeployment` | Deployment groups (EdgeFlow/CloudFlow rollouts) |
+| `CogniacDeploymentCapacityClass` | Deployment capacity classes |
+| `CogniacWorkflow` | Deployment workflows and versions |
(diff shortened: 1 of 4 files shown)


wskish (approved):
Reviewed: async/sync normalization parity confirmed equivalent, error-label mapping fills a real gap (all four labels are actually emitted in cli.py), and README claims back-check against the code. Ran the full smoke suite in a clean venv — 249 passed including the new tests. LGTM.

Optional non-blocker: async `detections()` itself isn't unit-tested (only `subjects()` got async coverage); a 4-line mirror would close it.

PR #165, opened 2026-06-16
retry _get/_head on 5xx+429; _post/_delete on 429 only (fixes #158)

bjw-0:
Fixes #158 (properly this time, per wskish's review on #162).

## What changed

wskish's review on #162 identified that the prior change was a no-op: `_get`/`_post`/`_delete` only retried on `CredentialError` (401 re-auth), so a `ServerError` raised by `raise_errors` propagated immediately regardless of the exception type assigned to 429.

This PR fixes the actual retry predicate on the data methods:

| Method | Before | After |
|---|---|---|
| `_get`, `_head` | 3 attempts, credential-only, no backoff | 8 attempts, exponential backoff, retries 5xx + 429 + connection errors + credential |
| `_post`, `_delete`, `_put` | 3 attempts, credential-only, no backoff | 8 attempts, exponential backoff, retries 429 + credential (NOT 5xx — avoids double-submit on non-idempotent ops) |

**Two new helpers in `common.py`:**
- `server_or_credential_error` — for idempotent reads (`_get`, `_head`)
- `rate_limit_or_credential_error` — for writes; retries 429 but not 5xx

**Three new end-to-end tests** (`test_coverage_smoke.py`):
- `test_get_retries_on_429_and_succeeds` — drives `_get` through a mocked 429→200 sequence; confirms 2 calls total
- `test_get_retries_on_5xx_and_succeeds` — drives `_get` through a mocked 500→200 sequence
- `test_post_retries_on_429_but_not_500` — confirms `_post` retries 429 but propagates 500 immediately

## Notes

- `#161` (the stacked commit) is already merged in master, so this branch is cleanly based on master.
- The CLI `--full-media` and `EdgeFlow.status()` docstring changes from the original `#162` are also already in master (merged via `#164`), so this PR is focused solely on the retry fix.
- 252 existing tests pass; no regressions.

Refs Cogniac/cogniac-sdk-py#158

### DIFF
--- a/cogniac/cogniac.py
+++ b/cogniac/cogniac.py
@@ -12,7 +12,7 @@ import re
 import httpx
 import sys
 from .common import retry, stop_after_attempt, wait_exponential, retry_if_exception
-from .common import server_error, raise_errors, CredentialError, credential_error
+from .common import server_error, raise_errors, CredentialError, credential_error, server_or_credential_error, rate_limit_or_credential_error
 from .credentials import stored_api_key, stored_url_prefix
 
 from .app     import CogniacApplication
@@ -248,7 +248,7 @@ class CogniacConnection(object):
         transport = httpx.HTTPTransport(retries=5)
         self.session = httpx.Client(transport=transport, headers=headers, follow_redirects=True)
 
-    @retry(stop=stop_after_attempt(3), retry=retry_if_exception(credential_error))
+    @retry(stop=stop_after_attempt(8), wait=wait_exponential(multiplier=0.5), retry=retry_if_exception(server_or_credential_error))
     def _head(self, url, timeout=None, **kwargs):
         """
         wrap requests session to re-authenticate on credential expiration
@@ -271,7 +271,7 @@ class CogniacConnection(object):
             raise
         return resp
 
-    @retry(stop=stop_after_attempt(3), retry=retry_if_exception(credential_error))
+    @retry(stop=stop_after_attempt(8), wait=wait_exponential(multiplier=0.5), retry=retry_if_exception(server_or_credential_error))
     def _get(self, url, timeout=None, **kwargs):
         """
         wrap httpx client to re-authenticate on credential expiration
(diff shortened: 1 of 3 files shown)


wskish (approved):
Approve. Verified the fix is real this time — `server_or_credential_error` / `rate_limit_or_credential_error` are now the actual retry predicates on the decorators (confirmed in cogniac.py:251–360), unlike the no-op in #162. The 429-safe-to-retry reasoning for writes is correct (a 429 means the request wasn't processed, so no double-submit risk), and the tests match the real plumbing (`_get` mocks `session.request()`, `_post` mocks `session.post()`). Ran all four new tests: 4 passed.

Non-blocking suggestions for a follow-up:
- **Bound the backoff.** `wait_exponential(multiplier=0.5)` over 7 retries waits ~63s worst case before a persistent 5xx/429 raises — a long silent hang for interactive calls. Consider `max=10`.
- **`Retry-After` is recorded but never honored** — `server_error`'s docstring implies waits use it, but `wait_exponential` ignores `exc.retry_after`. Either wire it in or trim the docstring.
- **`_delete`/`_put` retry paths are untested** — they share the predicate but go through `session.request()`, a different path than `_post`'s `session.post()`. A parametrized test would close the gap cheaply.
- Minor: docstring says "Non-idempotent methods (_post, _delete)" but `_put` gets the same treatment; and `server_or_credential_error` is defined above `server_error` it calls (works at call time, but reads oddly).

wskish (approved):
APPROVE (independent re-review). The retry asymmetry is textbook-correct: retry 5xx + 429 on idempotent GET/HEAD, but 429-only on POST/DELETE/PUT — a 429 means the write was rejected unprocessed (safe to retry), whereas a 5xx on a non-idempotent write has indeterminate server state (correctly NOT retried). Verified the predicates are genuinely wired (raise_errors runs inside the retried body, so ServerError/429 actually reach the tenacity predicate — unlike the no-op in #162), and the re-auth path is preserved across retries.

Non-blocking follow-ups: (1) `wait_exponential(multiplier=0.5)` has no `max`, summing to ~63s of silent waiting over 7 gaps before a persistent 5xx/429 finally raises — recommend `max=10`; (2) `Retry-After` is stashed on the exception but never consumed by the wait policy (docstring implies otherwise); (3) `_delete`/`_put` retry paths are untested (they go through `session.request()`, not `session.post()`).

PR #172, opened 2026-06-22
edgeflows status: add --start/--end time-range flags

wskish:
Closes #168

## Summary

`cogniac edgeflows status <gateway_id>` previously exposed only `--subsystem` and `--limit`, so any windowed analysis (e.g. throughput over the last 6 hours) required pulling a large `--limit` and filtering client-side. The SDK method `CogniacEdgeFlow.status()` already accepts `start=`/`end=`; this PR wires them through to the CLI.

## Changes

- Add `--start` / `--end` optional flags to the `edgeflows status` subparser and pass them through to `CogniacEdgeFlow.status(start=, end=)`. `--subsystem` and `--limit` continue to work alongside them.
- Update the help text and the `edgeflows status` flag listing in `CLAUDE.md`.
- Add smoke tests covering flag parsing and that the handler forwards `start`/`end` correctly.

## Accepted formats

Both flags use the repo's existing `_timestamp` arg type, so each accepts either:

- **epoch seconds** — e.g. `--start 1700000000`
- **ISO 8601 datetime** — e.g. `--end 2026-01-02T03:04:05Z`

They are converted to float epoch seconds (what `status()` expects). `--start` and `--end` may be given alone or together. No new dependency added (stdlib `datetime.fromisoformat` via the pre-existing helper).

## Verification

- `uv run pytest tests/test_coverage_smoke.py` — 254 passed.
- `cogniac edgeflows status --help` shows both new flags with `EPOCH_OR_ISO8601` metavar.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -173,6 +173,9 @@ cogniac media search            # search: --md5, --filename, --external-media-id
 cogniac edgeflows list          # list all edgeflows
 cogniac edgeflows get <id>      # get specific edgeflow
 cogniac edgeflows status <id>   # status events: --subsystem, --start, --end, --limit
+                                # --start/--end filter on the cloud-receipt timestamp (cc_timestamp),
+                                # not device time (clocks can diverge on a skewed/backfilling EdgeFlow);
+                                # --sort is not exposed (backend defaults to cloudcore_timestamp)
 cogniac cameras list            # list all cameras
 cogniac cameras get <id>        # get specific camera
 ```
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -172,7 +172,7 @@ cogniac media download <id>     # download media file to <id>.<ext>; use -o for 
 cogniac media search            # search: --md5, --filename, --external-media-id, --domain-unit, --limit
 cogniac edgeflows list          # list all edgeflows
 cogniac edgeflows get <id>      # get specific edgeflow
-cogniac edgeflows status <id>   # status events: --subsystem, --limit
+cogniac edgeflows status <id>   # status events: --subsystem, --start, --end, --limit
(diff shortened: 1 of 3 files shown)


wskish:
## Review: `--start`/`--end` for `edgeflows status`

**Verdict: approve-with-nits.**

**Wiring is correct.** `--start`/`--end` are registered on the `edgeflows status` subparser (dests `start`/`end`) and forwarded as `status(start=, end=)`. `--subsystem`/`--limit` still work; defaults are `None`, so omitted flags are skipped. The `_timestamp` arg type is genuinely pre-existing (no new dependency) and converts both epoch seconds and ISO 8601 into the float epoch the SDK expects. Tests follow repo conventions (parser-walk dest check + mocked-transport forwarding assertion) and pass.

**Should-fix (addressed in follow-up commit on this branch).** The help text didn't say which clock the window filters against. The flags filter on the cloud-receipt timestamp (`cc_timestamp`): the SDK `status()` defaults `sort=None` and sends no `sort` param, and the CLI exposes no `--sort`, so the backend schema applies its default `sort='cloudcore_timestamp'`, routing the query to filter on `cc_timestamp` rather than the device timestamp. That can differ from device time on a skewed/backfilling EdgeFlow. Clarified the `--start`/`--end` help text and the CLAUDE.md `edgeflows status` line accordingly.

**Nit.** The PR description mentions a README update that isn't in this PR's diff (only CLAUDE.md was touched) — harmless.

**Merge coordination.** PR #174 edits the same subparser arg-list, the same handler, and the same CLAUDE.md line — a guaranteed (but trivial) conflict. Recommend merging this PR (#172) first — it's smaller and closes #168 — then rebasing #174.

**Customer-data check: clean** (placeholders only).

PR #173, opened 2026-06-22
Fix cogstats aggregating zero for per-model detection subsystems

wskish:
Closes #167

## Root cause

`CogniacEdgeFlow.get_aggregated_stats()` (used by `cogstats`) queried:

```python
events = self.status(subsystem_name='model_detections*', start=..., end=..., sort='edgeflow_timestamp')
```

The `status` subsystem filter is **exact-match** — the trailing `*` is not a wildcard. Real detection subsystems are named `model_detections_<model_instance_id>` (one per deployed model, which is how all CloudFlows report telemetry), so `'model_detections*'` matched nothing and the aggregation loop iterated over zero events. `cogstats` therefore reported **0 detections / 0 pixels** for actively-processing devices.

## Fix

Self-contained client-side approach (no backend change):

- Pull all status events for the window (drop the broken exact-match filter) and aggregate over every event whose subsystem name starts with `model_detections_`.
- Derive the per-model id by stripping the `model_detections_` prefix instead of `split('_')[2]`, so model ids containing underscores stay intact.
- Skip unrelated subsystems and events with no matching status payload.
- Return shape and keys (`total`, `app`, `start_timestamp`, `end_timestamp`) are unchanged, so `cogstats` output is identical.

Applied to both the sync (`cogniac/edgeflow.py`) and async (`cogniac/async_edgeflow.py`) paths to preserve sync/async symmetry.

## Tests

Added no-creds regression tests (synthetic per-model `model_detections_<id>` events with placeholder ids):

- `tests/test_edgeflows.py` — asserts non-zero aggregation, per-model rollup, that unrelated subsystems are ignored, and that ids with underscores are handled; also asserts `get_aggregated_stats` no longer passes an exact-match subsystem filter.
- `tests/test_async_edgeflows.py` — async parity.

Full suite: `268 passed, 110 skipped` (live tests skipped — no creds), including `test_parity.py` (sync/async symmetry).

### DIFF
--- a/cogniac/async_edgeflow.py
+++ b/cogniac/async_edgeflow.py
@@ -227,17 +227,29 @@ class AsyncCogniacEdgeFlow(object):
         start = start - (start % REPORTING_PERIOD_SECONDS)
         end = end - (end % REPORTING_PERIOD_SECONDS)
 
+        # Detection telemetry is reported as one subsystem per deployed
+        # model, named ``model_detections_<model_instance_id>`` (this is how
+        # all CloudFlows report). The ``status`` subsystem filter is exact
+        # match only -- it does not support wildcards/prefixes -- so we pull
+        # all subsystems for the window and aggregate over every event whose
+        # subsystem name begins with ``model_detections_``.
+        MODEL_DETECTIONS_PREFIX = 'model_detections_'
+
         aggregated_stats = {'total': {}, 'app': {}}
         total_model_detections = 0
         total_aggregated_media_pixels = 0
         total_aggregated_gpu_pixels = 0
 
-        async for event in self.status(subsystem_name='model_detections*',
-                                       start=start,
+        async for event in self.status(start=start,
                                        end=end,
                                        sort='edgeflow_timestamp'):
-            app_id = event['subsystem'].split('_')[2]
+            subsystem = event.get('subsystem', '')
+            if not subsystem.startswith(MODEL_DETECTIONS_PREFIX):
+                continue
+            app_id = subsystem[len(MODEL_DETECTIONS_PREFIX):]
             event_status = event['status'].get(app_id)
+            if not event_status:
+                continue
             model_detections = event_status.get('model_detections', 0)
             media_pixels = event_status.get('aggregated_media_pixels', 0)
             gpu_pixels = event_status.get('aggregated_gpu_pixels', 0)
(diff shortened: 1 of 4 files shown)


wskish:
## Review — verdict: approve-with-nits

The fix is correct, sync/async-symmetric, and preserves the return shape so `bin/cogstats` output is unchanged. The prefix-strip `subsystem[len('model_detections_'):]` correctly handles model IDs containing underscores (the old `split('_')[2]` would have returned a truncated id). Regression tests are meaningful and pass.

### Root-cause refinement (cross-referenced with the #167 investigation)
The original `'model_detections*'` wasn't simply a no-op exact match. The backend **path route** (`/1/gateways/<gw>/status/<subsystem_name>`) actually *does* translate a trailing `*` into a DynamoDB `subsystem__beginswith` filter (`common/edgeflow.py:20-21`), and httpx sends `*` un-encoded — so that branch really is reached. It still returns 0 because, with `sort='edgeflow_timestamp'`, the query runs against the device-wide `gateway_id-gw_timestamp-index` partition and the prefix is applied as a **FilterExpression after reading only the first ~100-row page**. On a busy device those most-recent 100 rows are crowded with other subsystems, so the prefix matches get filtered out — whereas the exact `model_detections_<id>` query hits a dedicated per-subsystem partition where every row matches (hence the observed "0 vs 20"). See cogniac-sdk-py#167 and the backend issue edgeflow-core-api#85.

**This client-side scan-and-aggregate is the right fix precisely because it's independent of that filter-after-limit behavior** — it walks the full window rather than relying on the fragile prefix filter.

### Should-fix (suggestion — not yet applied)
`get_aggregated_stats()` now calls `self.status(start=, end=, sort=)` with **no `limit`**, so completeness relies entirely on the gateways/status endpoint emitting `paging.next` for `start/end/sort` queries. I confirmed against the backend that this endpoint *does* paginate and the `status()` generator drains `paging.next`, so the behavior is correct — but it's implicit. Worth a one-line code comment above the `self.status(...)` call in both `edgeflow.py` and `async_edgeflow.py` (e.g. "No limit: drain the full window via paging.next so the aggregation isn't truncated to one 100-row page") so a future reader doesn't reintroduce a cap. Comment-only, no logic change.

### Nit / follow-up (not blocking)
Per-app totals are gated on `if gpu_pixels:`, so a `model_detections > 0` event with zero/missing GPU pixels is silently dropped from the per-app rollup (pre-existing behavior, faithfully preserved here). Worth a separate issue to validate the per-window counter-summation semantics and this gpu-pixel gating.

**Customer-data check: clean** — diff, commit, and tests use placeholders only.

PR #174, opened 2026-06-22
edgeflows status: add --list-subsystems subsystem-discovery mode

wskish:
## Summary

`cogniac edgeflows status <gw>` defaults to `--limit 10`. A busy device emits high-frequency `model_detections_<id>` samples (one per model, every ~30s), which crowd the most-recent-N events — so low-frequency subsystems (`http-input-<app>`, `gpus`, `cpu`, `memory`, `connectivity`, `upload`, `event`) don't surface until `--limit ~2000`. A user following the natural "run with no filter to see what's reported, then query a specific subsystem" workflow never discovers them.

This adds option (1) from the issue: a cheap subsystem-discovery mode.

## New flag

`cogniac edgeflows status <gw> --list-subsystems`

- Scans a bounded window of device history **once** and emits the **distinct** `subsystem` values the device has reported — independent of per-subsystem sample frequency.
- The scan window is bounded by `--scan-limit` (default `2000`). The cap is **explicit and configurable**, never silent: if the scan hits `--scan-limit` (so a rarely-reported subsystem may still be missed), a diagnostic notice is written to **stderr** (`{"scan_capped": true, "scanned": N, "hint": ...}`), keeping stdout a clean JSON array — matching the existing `output()` stdout/stderr contract.

### Output shape

A JSON array, sorted by subsystem name, of:

```json
[
  {"subsystem": "cpu",                   "last_seen": 1718900000.0, "count": 1},
  {"subsystem": "gpus",                  "last_seen": 1718900001.0, "count": 1},
  {"subsystem": "http-input-bbbb",       "last_seen": 1718899990.0, "count": 1},
  {"subsystem": "model_detections_aaaa", "last_seen": 1718900160.0, "count": 3}
]
```

`last_seen` is the latest `edgeflow_timestamp` seen for that subsystem within the scanned window; `count` is how many times it appeared in the window.

## Unchanged behavior

The normal listing path (`--subsystem` / `--limit`, raw event listing) is untouched. The discovery code is localized to the `edgeflows status` handler.

## Verification

- Added non-live CLI tests (mock `status()` with synthetic events across several subsystems, placeholder ids `model_detections_aaaa` / `http-input-cccc`): assert the distinct set, sorted order, per-subsystem `count` + `last_seen`, that the scan-cap notice goes to **stderr** while stdout stays clean JSON, and that the no-flag path still lists raw events capped by `--limit`.
- `uv run pytest tests/` — 267 passed, 110 skipped (skips are live tests needing creds). Full smoke suite passes.

Closes #169

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -172,7 +172,7 @@ cogniac media download <id>     # download media file to <id>.<ext>; use -o for 
 cogniac media search            # search: --md5, --filename, --external-media-id, --domain-unit, --limit
 cogniac edgeflows list          # list all edgeflows
 cogniac edgeflows get <id>      # get specific edgeflow
-cogniac edgeflows status <id>   # status events: --subsystem, --limit
+cogniac edgeflows status <id>   # status events: --subsystem, --limit; --list-subsystems (+ --scan-limit) for distinct-subsystem discovery
 cogniac cameras list            # list all cameras
 cogniac cameras get <id>        # get specific camera
 ```
--- a/README.md
+++ b/README.md
@@ -137,6 +137,7 @@ cogniac subject media --subject-uid <uid> --full-media   # full media records, n
 cogniac media get --media-id <id>                  # media metadata
 cogniac media download --media-id <id> -o out.jpg  # download media to file
 cogniac edgeflow list                              # list edge devices
+cogniac edgeflow status --edgeflow-id <id> --list-subsystems   # discover which subsystems a device reports
 cogniac application create --body @app.json        # --body takes inline JSON, @FILE, or - (stdin)
 ```
 
--- a/cogniac/cli.py
+++ b/cogniac/cli.py
@@ -611,7 +611,7 @@ def _emit_subsystem_summary(edgeflow, args):
     stdout stays a clean JSON array; if the scan hits --scan-limit (so a rarer
     subsystem may still be missed), a diagnostic notice goes to stderr, per the
     CLI's stdout/stderr contract."""
-    scan_limit = getattr(args, 'scan_limit', None) or 2000
+    scan_limit = getattr(args, 'scan_limit', None) or 8000
     summary = {}
     scanned = 0
     for event in edgeflow.status(limit=scan_limit):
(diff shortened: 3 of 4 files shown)


wskish:
## Review: `--list-subsystems` discovery mode

**Verdict: approve-with-nits.**

The approach is correct. Client-side scan-and-dedupe is the right call here: the `GatewayStatus` table has no subsystem-only index, and DynamoDB has no `DISTINCT`, so there is no cheap server-side way to enumerate the distinct subsystems — scanning a bounded window of recent history and aggregating client-side is the pragmatic answer. The stdout/stderr contract is honored (pure-JSON on stdout, diagnostics on stderr), the normal `--subsystem`/`--limit` listing path is untouched, and the tests cover the distinct set, per-subsystem counts, `last_seen`, sort order, and the stderr scan-cap notice.

### Should-fix (addressed in a follow-up commit on this branch)

The default `--scan-limit` of `2000` was likely too small to actually deliver the feature's promise on a busy device. ~2000 events is only roughly 100 minutes of history on, say, a 10-model device, and that window can be made up entirely of high-frequency `model_detections_*` rows — exactly the crowding this mode exists to escape. Since each status page is capped at 100 events, a larger default just means more cheap paged GETs.

Follow-up commit raises the default to `8000` (in both the argparse flag default and the `_emit_subsystem_summary` fallback, kept consistent) and strengthens the `scan_capped` warning to make clear that on a busy device the entire scanned window may be high-frequency detection events, so a low-frequency subsystem can still be missing even when it is not actually rare — and that the user should raise `--scan-limit` to widen the window. Worth noting: that `scan_capped` stderr notice is the honesty mechanism that keeps this from being silently wrong, so it earns its keep.

### Nits

1. `last_seen` relies on the most-recent-first walk (`status()` defaults `reverse=True`), but the explicit max-comparison in the aggregation loop is good belt-and-suspenders against any future ordering change — keep it.
2. The README example uses the `--edgeflow-id` flag form while CLAUDE.md uses the positional `<id>` form. Both work via the alias layers, but tightening both to the canonical flag form would match convention.

### Merge coordination

PR #172 edits the same `edgeflows status` subparser arg-list, the same `cmd_edgeflows_status` handler, and the same CLAUDE.md line — a guaranteed (but trivial) conflict. Recommend merging #172 first, then rebasing this PR. The changes are additive and independent: the `--list-subsystems` early-return sits cleanly above the start/end listing path.

**Customer-data check: clean** (placeholders only).

PR #176, opened 2026-06-22
edgeflow status(): guarantee a top-level 'timestamp' on every record (#170)

wskish:
## Problem

`CogniacEdgeFlow.status()` (sync `cogniac/edgeflow.py`, async `cogniac/async_edgeflow.py`) yields raw backend status records, and these are **not uniform**: a significant fraction of records lack a top-level `timestamp` key, carrying only `gw_timestamp` and `cc_timestamp`. In a reported 7-day window, ~15.6% of samples (105 of 673) were missing `timestamp`.

Any client that sorts or diffs on `record["timestamp"]` -- the obvious thing to do for a time series -- raises `KeyError: 'timestamp'` partway through a result set. The non-uniformity is silent until a minority-schema record shows up, so short probes (`limit=1..10`) appear to work and longer windows fail.

## Fix

Combines options 1 and 2 from the issue:

1. **Guarantee a top-level `timestamp` on every yielded record.** When a record has no `timestamp`, alias it to `gw_timestamp` (the gateway-side sample clock, present on every record). An existing `timestamp` is never overwritten; `gw_timestamp` and `cc_timestamp` are left intact. Applied identically to the sync and async generators (the dict is mutated in place before `yield`, which is safe since the generator already yields fresh dicts from the JSON response).

2. **Document the record timestamp schema** in both `status()` docstrings:
   - `gw_timestamp` -- gateway-side sample clock; always present. Canonical for time-series math / differencing cumulative counters, and **recommended for rate math**.
   - `cc_timestamp` -- CloudCore receive clock; always present.
   - `timestamp` -- now guaranteed on every record (aliased to `gw_timestamp` when the backend omits it).

## Tests

Adds sync and async parity tests that mock the connection to return a paging envelope whose `data` mixes records with and without `timestamp`. They assert that every yielded record has a `timestamp`, that records already carrying one keep their original value (not clobbered by `gw_timestamp`), and that a record missing one gets `gw_timestamp`'s value.

`uv run pytest tests/test_edgeflows.py tests/test_async_edgeflows.py tests/test_parity.py tests/test_coverage_smoke.py` -> 268 passed, 11 skipped (live-creds tests).

## Related

Complements #175, which client-side-works-around this same non-uniformity in `--list-subsystems` (`last_seen` reads `cc_timestamp`/`timestamp`/`gw_timestamp` in turn). This PR fixes the root cause at the SDK boundary so callers no longer need to.

Closes #170

### DIFF
--- a/cogniac/async_edgeflow.py
+++ b/cogniac/async_edgeflow.py
@@ -170,6 +170,16 @@ class AsyncCogniacEdgeFlow(object):
         reverse (bool)         reverse the sorting order: sort high to low
         limit (int)            yield maximum of limit results
         sort (str)             optional sort field
+
+        Record timestamp schema (each yielded record):
+          gw_timestamp  gateway-side sample clock; always present. This is the
+                        canonical clock for time-series math / differencing
+                        cumulative counters, and is recommended for rate math.
+          cc_timestamp  CloudCore receive clock; always present.
+          timestamp     now guaranteed on every yielded record. The backend
+                        omits it on a minority of records; in that case it is
+                        aliased to gw_timestamp here. An existing timestamp is
+                        never overwritten.
         """
         args = []
         if start is not None:
@@ -201,6 +211,13 @@ class AsyncCogniacEdgeFlow(object):
         while url:
             resp = await get_next(url)
             for data in resp['data']:
+                # Guarantee a top-level 'timestamp' on every record. The
+                # backend omits it on a minority of records (they carry only
+                # gw_timestamp/cc_timestamp); alias to gw_timestamp so callers
+                # can sort/diff on data['timestamp'] uniformly. Never clobber
+                # an existing timestamp.
+                if 'timestamp' not in data and 'gw_timestamp' in data:
+                    data['timestamp'] = data['gw_timestamp']
                 yield data
                 count += 1
                 if limit and count == limit:
--- a/cogniac/edgeflow.py
+++ b/cogniac/edgeflow.py
@@ -354,6 +354,16 @@ class CogniacEdgeFlow(object):
                                (seconds since epoch)
         reverse (bool)         reverse the sorting order: sort high to low
         limit (int)            yield maximum of limit results
+
+        Record timestamp schema (each yielded record):
+          gw_timestamp  gateway-side sample clock; always present. This is the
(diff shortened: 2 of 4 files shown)


wskish:
## Review — verdict: approve-with-nits

Correct, minimal, sync/async-symmetric, well-tested, no customer data. The alias is applied inside the per-page loop before `yield`, so it covers every record across pagination; it uses a membership test (`'timestamp' not in data`) so a legitimately-falsy `timestamp` (e.g. `0.0`) is preserved; and it leaves `cc_timestamp`/`gw_timestamp` intact. In-place mutation is safe (each record is a fresh `resp.json()` dict). Tests cover all four cases (uniform timestamp, no-clobber, missing→gw_timestamp, sort-no-longer-raises) with async parity.

### Nits (non-blocking)

1. **"Canonical" clock wording vs #175.** This docstring calls `gw_timestamp` canonical, while the already-merged #175 (`_emit_subsystem_summary`) calls `cc_timestamp` canonical and prefers it in its `last_seen` fallback. Both are defensible — `gw_timestamp` is the sample clock (right denominator for differencing cumulative counters), `cc_timestamp` is the cloud-receipt clock (what `--start/--end` filter on). Suggest the docstring distinguish *canonical for rate math* (gw) from *canonical for cloud-side ordering/filtering* (cc) so a reader of both isn't misled. Wording only — behavior is fine. This does **not** make #175's fallback redundant (it reads `cc_timestamp` first), and they don't collide.

2. **Aliased-vs-real `timestamp` semantics (unverifiable from the SDK).** On the ~84% of records that already carry `timestamp`, is the backend's `timestamp` equal to `gw_timestamp` or closer to `cc_timestamp`? If the latter, aliasing the missing ~16% to `gw_timestamp` makes those subtly inconsistent with the majority. The issue's framing (gw always present) makes gw the safe pick, but worth a one-line confirmation from whoever owns the gateway status producer that backend `timestamp ≈ gw_timestamp`.

### Merge ordering
#176 and #177 both append a test block to `tests/test_edgeflows.py` at the same anchor → near-certain trivial conflict. The source edits don't overlap (`status()` vs `metrics()`/`all_metrics()`). Recommend merging **#176 first**, then rebasing #177 (stack both test blocks).

PR #177, opened 2026-06-22
Fix `edgeflows metrics list`: wire SDK params and CLI flags (#171)

wskish:
## Problem

`cogniac edgeflows metrics list` always returned a 400. The backend reported `metric_name`, `tenant_id`, and `ef_id` all missing even when `--tenant`/`--edgeflow-id` were supplied. The subcommand was half-wired across two layers.

## Two layered bugs

**SDK (`cogniac/edgeflow.py`)**
- `metrics()` called `GET /1/metrics/ef` with `gateway_id`, but per the `ef-metrics-api` contract that endpoint requires `ef_id` (not `gateway_id`) plus `tenant_id`. The wrong param name was sent and `tenant_id` was never injected.
- `all_metrics()` called `GET /1/metrics`, which requires `metric_name` + `tenant_id`, but injected no `tenant_id`.

**CLI (`cogniac/cli.py`)**
- The `metrics list` subparser exposed only `--edgeflow-id` — there was no `--metric-name`, so the required param could not be passed.
- `cmd_edgeflow_metrics_list` invoked `ef.metrics()` / `all_metrics(cc)` with no params, forwarding nothing.

## Fix

**SDK**
- `metrics()` now injects `ef_id=self.gateway_id` and `tenant_id` from the connection via `setdefault` (explicit caller values still win), and passes arbitrary `**params` (`metric_name`, `start`, `end`) through.
- `all_metrics()` now injects `tenant_id` via `setdefault`.
- Docstrings updated to match the contract.

**CLI**
- `metrics list` gains `--metric-name` (required) and `--start`/`--end` (epoch seconds or ISO 8601, via the shared `_timestamp` arg type used by `edgeflows status`). `--edgeflow-id` stays optional.
- The handler builds the params dict, requires `--start`/`--end` to be supplied together (clear error otherwise, since the API rejects an unpaired bound), and forwards to `ef.metrics(**params)` when scoped to one device, else `CogniacEdgeFlow.all_metrics(cc, **params)`. `tenant_id`/`ef_id` come from the SDK; `--tenant` flows into the connection's tenant.

## Tests

Mocked-transport coverage in `tests/test_edgeflows.py`: asserts `metrics()` sends `ef_id`+`tenant_id` (not `gateway_id`), `all_metrics()` sends `tenant_id`, the CLI forwards `metric_name` and paired `start`/`end`, and an unpaired `--start` errors before any request is issued. Updated the alias-routing pair in `tests/test_coverage_smoke.py` for the now-required flag.

`uv run pytest tests/test_edgeflows.py tests/test_coverage_smoke.py tests/test_parity.py` → 267 passed, 7 skipped (live-only). No live API calls.

Closes #171


wskish:
## Review — verdict: request-changes

The sync-side fix is correct and faithful to the `ef-metrics-api` contract — verified: `metrics()` now sends `ef_id=self.gateway_id` + injects `tenant_id` via `setdefault` (so explicit callers win); `all_metrics()` injects `tenant_id`; `connection.tenant_id` is the right attribute (raw string, not the `tenant` property that calls `_require_tenant()`); `--metric-name` is `required=True`; `--start/--end` use the shared `_timestamp` type with both-or-neither enforced before any request; `_reg_ef_metrics` registers the new dests for both the nested command and the `edgeflow-metrics` flat alias; no other caller passes the old `gateway_id` param. Clean of customer data.

One blocking item and a couple of smaller ones:

### Blocking

**Async counterparts still carry both bugs.** The fix landed only in `cogniac/edgeflow.py`. In `cogniac/async_edgeflow.py`:
- `metrics()` still does `params.setdefault('gateway_id', self.gateway_id)` and never injects `tenant_id` → `/1/metrics/ef` (requires `ef_id` + `tenant_id`) will 400 exactly as the sync path did.
- `all_metrics()` never injects `tenant_id` → `/1/metrics` 400s.
- Docstrings still say "identified via the gateway_id query parameter."

CLAUDE.md requires sync/async symmetry, and the smoke test only checks signatures, not behavior, so this wouldn't be caught automatically. Mirror the sync fix: send `ef_id=self.gateway_id`, `setdefault('tenant_id', connection.tenant_id)` (`AsyncCogniacConnection.tenant_id` is a raw attribute), and update the docstrings.

### Should-fix

**Test drains capsys twice.** In `test_cli_metrics_list_rejects_unpaired_start`:
```python
err = capsys.readouterr().out + capsys.readouterr().err
```
`readouterr()` resets the buffers, so the second call returns empty and `err` never sees stderr; the variable is also never asserted. Capture once (`captured = capsys.readouterr()`) and assert the error message, or drop the dead line (the `conn.last_get is None` assertion does carry the test).

### Nit

`error_exit("UsageError", …)` isn't a registered `_ERROR_TYPES` key, so it maps to the generic `'error'` type. Functionally fine (clear message, stderr, exit 1); `"ClientError"` would be marginally more consistent. (`int(start)` truncating the `_timestamp` float is correct — the API wants integer epoch seconds.)

### Merge ordering
Conflicts with #176 in `tests/test_edgeflows.py` (same append anchor) are near-certain but trivial. Recommend merging #176 first, then rebasing this PR and stacking both test blocks.

wskish:
Follow-up commit `c785cf2` addresses the review:

- **Async parity (blocking):** mirrored the #171 fix into `cogniac/async_edgeflow.py`. `metrics()` now sends `ef_id=self.gateway_id` (not `gateway_id`) and `setdefault('tenant_id', self._cc.tenant_id)`; `all_metrics()` now `setdefault('tenant_id', connection.tenant_id)`. Both docstrings updated to match the sync versions. The sync/async `metrics()`/`all_metrics()` bodies are now equivalent modulo async/await (plus the async retry decorator).
- **capsys double-read test fix:** `test_cli_metrics_list_rejects_unpaired_start` now captures `capsys.readouterr()` once and asserts the unpaired `--start`/`--end` error message is actually present (in addition to `conn.last_get is None`).
- **Error-type nit:** the unpaired start/end guard in `cmd_edgeflow_metrics_list` now uses the registered `ClientError` type (consistent with the function's other guard) instead of `UsageError`; message unchanged.
- Added async param-forwarding tests in `tests/test_async_edgeflows.py` locking the symmetry.

All of `test_edgeflows.py test_async_edgeflows.py test_coverage_smoke.py test_parity.py` pass (270 passed, 11 skipped live-cred tests).

PR #186, opened 2026-07-15
Add deployment deploy/deploy-status verbs; warn that setting target does not deploy

wskish:
## Summary

`cogniac deployment target workflow set` (SDK `set_target_workflow`) records `target_workflow_id` via `POST /1/deploymentGroups/{id}/targetWorkflow` but **never dispatches a rollout** — and no CLI verb or public SDK method did. During a production revert this looked like a completed deploy while `current_workflow_id` was unchanged on every group. The real deploy mechanism (`POST /1/deploymentGroups/{id}` with `next_workflow_id`, or `deploy_now_workflow_id` for a schedule-bypassing immediate dispatch) was only reachable via the private `_post`, and its long synchronous server-side dispatch could time out client-side with no way to tell success from failure.

## Changes

**SDK** (sync `CogniacDeployment` + async `AsyncCogniacDeployment`):
- `deploy(workflow_id, now=False, timeout=None)` — dispatches a rollout via `POST /1/deploymentGroups/{id}` with `next_workflow_id` (immediate when the group has no `scheduled_time`, otherwise picked up by the scheduler) or `deploy_now_workflow_id` with `now=True`. The server blocks until every EdgeFlow in the group accepts, so the call takes a configurable read timeout with a generous default (`DEPLOY_DEFAULT_TIMEOUT`, 300s) instead of the connection's 60s. Returns the updated group JSON.
- `deploy_status()` — idempotent convergence check: re-GETs the group and returns `target/current/next/deploy_now` workflow ids plus a `converged` boolean (`current == target` and no pending `next`). Resolves the ambiguity when a long dispatch times out client-side but completes server-side.
- `set_target_workflow` docstrings now state they record intent only.

**CLI**:
- `cogniac deployment deploy --deployment-group-id <id> --workflow-id <wf> [--now] [--timeout SECONDS]`
- `cogniac deployment deploy-status --deployment-group-id <id>`
- `deployment target workflow set` now emits a **stderr** warning (stdout stays pure JSON): target set but NOT deployed, with the exact deploy command to run; its help text now says it records the target without deploying.
- A client read timeout during deploy prints a structured stderr error (`type: timeout`) explaining the dispatch may still complete server-side, with a `deploy-status` hint, and exits nonzero instead of a bare traceback.

**Docs**: README CLI section gains the new verbs and the record-only caveat.

## Tests

New `tests/test_deployment.py` (mocked transport, no credentials, no live API calls, nothing dispatched): deploy request body for both modes, endpoint, timeout plumbing and default, `deploy_status` convergence truth table, CLI wiring for deploy/deploy-status, the read-timeout guidance path, and the stderr warning on target set (stdout stays parseable JSON). Smoke-test method lists updated so the sync/async symmetry check covers the new methods.

```
304 passed, 110 skipped (live tests skipped — no credentials)
```

Fixes #180
Fixes #185

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,13 @@ cogniac media download --media-id <id> -o out.jpg  # download media to file
 cogniac edgeflow list                              # list edge devices
 cogniac edgeflow status --edgeflow-id <id> --list-subsystems   # discover which subsystems a device reports
 cogniac application create --body @app.json        # --body takes inline JSON, @FILE, or - (stdin)
+
+cogniac deployment deploy --deployment-group-id <id> --workflow-id <wf>   # DISPATCH a workflow rollout
+                                                   # (--now bypasses the group schedule; --timeout raises the
+                                                   #  client read timeout — the server blocks until every
+                                                   #  EdgeFlow accepts, default 300s)
+cogniac deployment deploy-status --deployment-group-id <id>               # rollout convergence status
+cogniac deployment target workflow set ...         # records target_workflow_id only — does NOT deploy
 ```
 
 Run `cogniac <noun> --help` to explore the tree interactively, or `cogniac commands` for the full machine-readable catalog. An unknown command suggests the closest match.
--- a/cogniac/async_deployment.py
+++ b/cogniac/async_deployment.py
@@ -5,6 +5,7 @@ Copyright (C) 2024 Cogniac Corporation
 """
 
 from .common import retry, stop_after_attempt, wait_exponential, retry_if_exception, server_error
+from .deployment import DEPLOY_DEFAULT_TIMEOUT
 
 
 class AsyncCogniacDeployment(object):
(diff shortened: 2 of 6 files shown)


wskish (commented):
Reviewed the diff, ran the suite locally, and live-verified the read-only surface.

**Correctness**
- `deploy()` POSTs to `/1/deploymentGroups/{id}` with `next_workflow_id` (or `deploy_now_workflow_id` with `now=True`) — matches the real dispatch mechanism, and the `timeout=` kwarg is properly plumbed through `CogniacConnection._post`.
- `deploy_status()` convergence (`current == target` and no pending `next`, with a not-None guard) is right, and it is a pure GET.
- Verified the safety property that matters most here: `httpx.ReadTimeout` is not matched by the `server_error` retry predicate, so the method-level `@retry` will not re-dispatch on a client read timeout; the CLI's structured stderr timeout path then applies.
- The stderr warning after `target workflow set` keeps stdout as pure JSON (confirmed by the test).

**Non-blocking note**: the method-level `@retry(retry=retry_if_exception(server_error))` on `deploy()` will re-POST the dispatch on a 5xx or connect error (up to 8 attempts). That is consistent with the existing `set_target_workflow`/`prepull_start` pattern, but `common.py`'s own guidance says non-idempotent POSTs should retry on rate-limit/credential errors only, and a double dispatch is more consequential for this verb than for the others. Worth either switching the predicate or leaving a comment saying the retry is intentional.

**Tests**: checked out the branch and ran the suite locally — 304 passed, 110 skipped. New tests are mock-only, no live calls, placeholders throughout.

**Live verification (read-only)**: on a live tenant, `deployment deploy-status` correctly reported `converged: true` with `current == target` and `next null` on two deployment groups. Did not exercise the mutating `deploy` verb live; its request shape is covered by the mocked tests.

Public-repo hygiene: diff, commit message, and PR body contain no customer-identifying information. LGTM apart from the retry-predicate note.

wskish:
Addressed the review finding in ee18076: removed the blanket `@retry(server_error)` decorator from `deploy()` (sync and async). The dispatch POST is not idempotent — a 5xx after the server began processing could double-dispatch the rollout — and `common.py`'s own predicate guidance excludes 5xx retry on non-idempotent methods. The transport's `_post` still retries 429 rate-limits and credential refresh (rejected before processing, so safe), and `deploy_status()` keeps its retry as an idempotent GET.

Docstrings now state the call is not auto-retried and direct callers to `deploy_status()` after any failure before retrying. Added tests: a 5xx propagates from `deploy()` after exactly one POST attempt, a read timeout propagates un-retried (the CLI's timeout-guidance path is unaffected), and `deploy_status()` still retries on a transient 5xx. Full suite: 307 passed, 110 skipped (live tests, no credentials).

PR #187, opened 2026-07-15
Add workflow version enumeration: SDK get_all_versions + CLI 'workflow version list'

wskish:
## Summary

Fixes #181 — there was no way to enumerate workflow versions; the only options were `workflow get <BASE:version>` (one id at a time) and `workflow list` (tenant bases only), so discovering versions meant guessing `BASE:N` ids until 404.

The backend already exposes `GET /1/workflows/{base_id}/versions` — a paginated, newest-first summary list (`{"data": [...], "last_key": <cursor>}` with `reverse`/`limit`/`last_key` query params). This PR surfaces it in the SDK and CLI.

## SDK

- `CogniacWorkflow.get_all_versions(connection, base_id, reverse=True, limit=None, last_key=None)` — a generator that drains the `last_key` cursor and yields `CogniacWorkflow` objects, mirroring `CogniacDeployment.history`.
- `AsyncCogniacWorkflow.get_all_versions(...)` — async generator, same signature (sync/async symmetry).
- A full `<base_id>:<version>` id is accepted; the version suffix is stripped, matching the backend's own tolerance on version-scoped reads — handy when all you have is the workflow id off a deployment group.

## CLI

```
cogniac workflow version list --base-id BASE [--limit N] [--reverse|--no-reverse]
cogniac workflow version list BASE                 # deprecated positional mirror
cogniac workflow list --base-id BASE               # --base flag form proposed in the issue
```

JSON output includes every field the list endpoint returns (`workflow_id`, `version`, `name`, `created_at`, `created_by`, ...); `--format table` uses new `workflow_version` columns. The hidden flat `workflow-version list` alias is registered via the shared registrar, and `workflow list` also gains `--limit`.

## Tests

- Smoke: method existence sync+async, generator check, and mocked-transport pagination — empty envelope, `last_key` drain across pages, mid-page `--limit` stop without following the cursor, and `BASE:version` truncation.
- Live (read-only, credential-gated): lists versions of the tenant's first workflow base.

`pytest tests/ -q`: **295 passed, 111 skipped** (skips are the credential-gated live tests). Also verified end-to-end against a live tenant: `workflow list`, `workflow version list --base-id ...`, `workflow list --base-id BASE:5 --limit 3`, and the flat alias all return clean JSON arrays.

### DIFF
--- a/cogniac/workflow.py
+++ b/cogniac/workflow.py
@@ -108,6 +108,49 @@ class CogniacWorkflow(object):
         return CogniacWorkflow(connection, resp.json())
 
     ##
+    #  get_all_versions
+    ##
+    @classmethod
+    def get_all_versions(cls, connection, base_id, reverse=True, limit=None, last_key=None):
+        """
+        Yield every version of a workflow base as CogniacWorkflow objects,
+        following the DynamoDB last_key cursor until the versions are drained.
+
+        base_id (str)    the workflow base id; a full workflow_id of the form
+                         <base_id>:<version> is also accepted (the version
+                         suffix is ignored)
+        reverse (bool)   newest first when True (default)
+        limit (int)      yield a maximum of limit versions
+        last_key (str)   resume from a previous last_key cursor
+
+        See GET /1/workflows/{base_id}/versions.
+        """
+        base_id = base_id.split(':', 1)[0]  # tolerate a full <base_id>:<version>
+
+        @retry(stop=stop_after_attempt(8), wait=wait_exponential(multiplier=0.5), retry=retry_if_exception(server_error))
+        def get_next(last_key):
+            params = {'reverse': reverse}
+            if limit is not None:
+                params['limit'] = limit
+            if last_key is not None:
+                params['last_key'] = last_key
+            resp = connection._get("/1/workflows/%s/versions" % base_id, params=params)
+            return resp.json()
+
+        count = 0
+        while True:
+            resp = get_next(last_key)
+            data = resp['data'] if isinstance(resp, dict) and 'data' in resp else resp
+            for record in data or []:
+                yield CogniacWorkflow(connection, record)
+                count += 1
+                if limit and count == limit:
+                    return
+            last_key = resp.get('last_key') if isinstance(resp, dict) else None
+            if not last_key:
+                return
+
+    ##
     #  get_version
     ##
     @classmethod
--- a/cogniac/workflow.py
+++ b/cogniac/workflow.py
@@ -116,6 +116,11 @@ class CogniacWorkflow(object):
         Yield every version of a workflow base as CogniacWorkflow objects,
         following the DynamoDB last_key cursor until the versions are drained.
 
+        The versions endpoint yields summary records (workflow_id, version,
+        name, created_at, created_by, description, edgeflow_model, base_id,
+        tenant_id) WITHOUT app_specs; fetch the full workflow via get()
+        before diffing or summarizing.
+
         base_id (str)    the workflow base id; a full workflow_id of the form
                          <base_id>:<version> is also accepted (the version
                          suffix is ignored)
--- a/cogniac/async_workflow.py
+++ b/cogniac/async_workflow.py
@@ -83,6 +83,11 @@ class AsyncCogniacWorkflow(object):
         AsyncCogniacWorkflow objects, following the DynamoDB last_key cursor
         until the versions are drained.
 
+        The versions endpoint yields summary records (workflow_id, version,
+        name, created_at, created_by, description, edgeflow_model, base_id,
+        tenant_id) WITHOUT app_specs; fetch the full workflow via get()
+        before diffing or summarizing.
+
         base_id (str)    the workflow base id; a full workflow_id of the form
                          <base_id>:<version> is also accepted (the version
                          suffix is ignored)
--- a/cogniac/async_workflow.py
+++ b/cogniac/async_workflow.py
@@ -77,6 +77,47 @@ class AsyncCogniacWorkflow(object):
         return AsyncCogniacWorkflow(connection, resp.json())
 
     @classmethod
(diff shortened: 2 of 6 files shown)


wskish (commented):
Reviewed the diff, ran the suite locally, and live-verified against a live tenant.

**Correctness**: `get_all_versions` correctly follows the `last_key` cursor (mirroring `CogniacDeployment.history`), stops mid-page on `limit` without following the cursor, tolerates an empty `data` envelope, and the `BASE:version` suffix truncation works as documented. Sync/async symmetry is maintained and the smoke-test generator check covers both.

**Convention fit**: retry decorators, docstring style (`See GET ...`), CLI verb registration via `_add_verb`/`_reg_wf_version` with the flat alias, and `workflow_version` table columns all match existing patterns.

**Tests**: checked out the branch and ran the suite locally — 295 passed, 111 skipped. Mocked pagination tests use invented placeholders only; the live test is properly credential-gated and read-only.

**Live verification (read-only)**: on a live tenant, `workflow list --base-id BASE` enumerated all versions of a workflow base (33 versions, newest first, `created_at`/`created_by` populated); `workflow version list` with `--limit`, `--no-reverse`, and a full `BASE:version` id all behaved as documented.

Minor, non-blocking: `workflow list --limit` without `--base-id` fetches the full tenant list and slices client-side — fine for the current endpoint, just noting it isn't server-side pagination.

cogniac/workflow.py
```diff
@@ -107,6 +107,49 @@ def new_version(cls, connection, base_id, body):
...
+        following the DynamoDB last_key cursor until the versions are drained.
+
+        base_id (str)    the workflow base id; a full workflow_id of the form
+                         <base_id>:<version> is also accepted (the version
+                         suffix is ignored)
+        reverse (bool)   newest first when True (default)
+        limit (int)      yield a maximum of limit versions
+        last_key (str)   resume from a previous last_key cursor
+
+        See GET /1/workflows/{base_id}/versions.
```
spatel-cog, line 131: **Non-blocking, but worth a docstring line here**: the versions endpoint returns summary records -- on the production tenant I verified, every record carries exactly 9 fields (`workflow_id`, `version`, `name`, `created_at`, `created_by`, `description`, `edgeflow_model`, `base_id`, `tenant_id`) and **no `app_specs`**.

`get_all_versions` wraps these in full `CogniacWorkflow` objects, which sets up a silent footgun once #188 lands: calling `.diff()` / `.summary()` (or the module functions) on version-list objects quietly reports zero apps. A diff between two summary records shows only name/description changes with empty `apps_changed`; a diff of a summary record against a full `get()` result reports every app as added. Nothing crashes -- the output is just wrong in a way that looks plausible.

Two cheap mitigations:

1. Here: one docstring line -- "yields summary records without `app_specs`; fetch the full workflow via `get()` before diffing or summarizing."
2. In #188 (or a follow-up): have `workflow_diff` / `workflow_summary` distinguish the `app_specs` key being absent from an empty list, and emit a caveat field (or raise) when absent -- "no key" means "summary record," not "workflow with no apps."

spatel-cog (commented):
Review by Claude, on behalf of @spatel-cog. I checked out the branch and live-verified it read-only against the same deep-history production tenant as my #188 review (a workflow base at 300+ versions). Identifiers scrubbed as usual.

**Live validation:**

- **Full drain at scale**: enumerated all 300+ versions of the base in ~5s; versions strictly monotonic newest-first, so the cursor-following logic is correct across many pages.
- **`--no-reverse` genuinely works server-side**: returned versions 0, 1, 2 -- worth calling out because `reverse` is passed as a Python bool query param, and this confirms the serialization that the `CogniacDeployment.history` precedent relies on also behaves correctly for `False`.
- **Suffix truncation**: `workflow list --base-id BASE:5 --limit 3` returned the newest 3 versions of the base, as documented. The positional mirror and the flat alias parse as described (`_id` registers both the positional and flag forms).

**Implementation**: a near-verbatim mirror of `CogniacDeployment.history` (params handling, envelope tolerance, cursor loop, mid-page limit stop) -- and slightly more defensive than its precedent: `for record in data or []` where `history` would crash on a `data: null` envelope. The mocked pagination tests match what I observed live, and `cmd_workflows_list` gaining a `ClientError` handler it previously lacked is a nice incidental fix.

**One substantive but non-blocking note** -- the versions endpoint returns summary records with no `app_specs`, which sets up a silent footgun with #188's `diff`/`summary`; details and suggested mitigations on the inline comment at `get_all_versions`.

**Merge sequencing with #188**: the two branches textually collide in three files (`_WORKFLOW_METHODS` in `test_coverage_smoke.py`, the workflow verb-registration block in `cli.py`, and `workflow.py`). All trivial, but whichever lands second needs a rebase, and the merged `_WORKFLOW_METHODS` should carry all three additions (`get_all_versions`, `diff`, `summary`).

**Minor**: after the first page, subsequent requests re-send the original `limit` rather than the remainder -- bounded over-fetch since the client stops mid-page anyway; inherited from the `history` precedent, fine as-is. Agreed with the earlier review that the client-side slice on `workflow list --limit` (without `--base-id`) is acceptable for the current endpoint.

**Verdict: LGTM.** Correctness verified live at scale, the implementation follows and slightly improves on the established pagination pattern, and the tests cover the envelope edge cases. The docstring caveat below is worth adding but does not block.

cogniac/workflow.py
wskish: Landed in e9e31fa: both the sync and async `get_all_versions` docstrings now state that the versions endpoint yields summary records (workflow_id, version, name, created_at, created_by, description, edgeflow_model, base_id, tenant_id) without `app_specs`, and to fetch the full workflow via `get()` before diffing or summarizing. The CLI help for `workflow version list` and `workflow list --base-id` carries the same caveat. The absent-key-vs-empty-list distinction in `workflow_diff`/`workflow_summary` is left for #188 or a follow-up as suggested.

spatel-cog (approved):

PR #188, opened 2026-07-15
Add workflow diff and model-composition summary

wskish:
## Summary

Fixes #182.

There was no affordance to see what changed between two workflow versions: `workflows get` returns a 100+ entry `app_specs` array where `name`/`type` are often null, and the entire change between versions can be a single app's `model_runtime_image` retag — nearly invisible in raw JSON. This adds a pure client-side diff and a model-composition summary (no new API endpoints).

## CLI

```
cogniac workflow diff <workflow_a> <workflow_b>
cogniac workflow summary --workflow-id <workflow_id>
```

`diff` fetches both workflows and emits a compact, reviewable JSON structure:

- `apps_changed` — keyed by `application_id`; `model_runtime_image` and threshold changes surfaced prominently as `{"old", "new"}`, plus a generic dotted-path deep diff of the whole app spec (`changed`), so any other field change is caught too
- `apps_added` / `apps_removed` — compact summary rows
- `top_level_changes` — scalars (name, description, edgeflow_model, ...) as `old`/`new`; composite tenant config snapshots reported tersely as changed-path counts
- `identical: true/false`

`summary` emits one row per app spec — `application_id`, `model_runtime_image`, threshold fields (and name/type when the spec actually carries them) — so a human can read the model composition without hand-joining against `apps get`.

Both handle `model_runtime_image` and threshold-like fields at the top level of each app spec **and** nested under `app_type_config`.

Example `diff` output (fixture data):

```json
{
  "workflow_a": {"workflow_id": "wfbase0001:3", "name": "example-inspection-pipeline", "version": 3},
  "workflow_b": {"workflow_id": "wfbase0001:4", "name": "example-inspection-pipeline-v2", "version": 4},
  "top_level_changes": {"name": {"old": "example-inspection-pipeline", "new": "example-inspection-pipeline-v2"}},
  "apps_added": [{"application_id": "appcount000001", "model_runtime_image": "registry.example.com/runtime/counter:0.1.0"}],
  "apps_removed": [{"application_id": "appocr00000001", "model_runtime_image": "registry.example.com/runtime/ocr:7.8.9"}],
  "apps_changed": {
    "appdetect00001": {
      "model_runtime_image": {"old": "registry.example.com/runtime/detector:1.2.3", "new": "registry.example.com/runtime/detector:1.2.4"},
      "changed": {"model_runtime_image": {"old": "...:1.2.3", "new": "...:1.2.4"}}
    }
  },
  "identical": false
}
```

## SDK

The logic lives in reusable module-level functions, not in CLI parsing:

- `cogniac.workflow_diff(workflow_a, workflow_b)` and `cogniac.workflow_summary(workflow)` — accept `CogniacWorkflow`/`AsyncCogniacWorkflow` instances or plain workflow dicts; exported from the package root
- `CogniacWorkflow.diff(other)` / `.summary()` instance methods, mirrored on `AsyncCogniacWorkflow` (pure local computation, not coroutines)

## Tests

`tests/test_workflow_diff.py` (20 tests, no live API, invented placeholder ids): image retag (top-level and nested under `app_type_config`), threshold change (both locations), added/removed apps, unchanged apps not reported, top-level scalar and terse composite changes, identical workflows, object/dict inputs, package exports, missing `app_specs` tolerated, and CLI parser wiring. The workflow method-pairing smoke test now also covers `diff`/`summary`.

```
310 passed, 110 skipped (live-credential tests)
```

### DIFF
--- a/cogniac/workflow.py
+++ b/cogniac/workflow.py
@@ -8,6 +8,255 @@ from .common import retry, stop_after_attempt, wait_exponential, retry_if_except
 
 
 ##
+#  pure helpers: workflow summary / diff
+##
+
+# Top-level workflow fields excluded from the diff: identity fields (reported in
+# the diff header), per-version bookkeeping, and the bulky rendered deployment
+# artifact whose differences are derived from app_specs anyway.
+_DIFF_EXCLUDED_TOP_LEVEL = frozenset([
+    'app_specs', '_apply_jsons',
+    'workflow_id', 'base_id', 'version',
+    'created_at', 'created_by', 'updated_at', 'updated_by',
+    'latest_version', 'latest_workflow_id',
+    'system_software_updates_available',
+])
+
+_MISSING = object()
+
+
+def _as_workflow_dict(workflow):
+    """Accept a CogniacWorkflow / AsyncCogniacWorkflow instance or a plain
+    workflow dict and return a plain dict of the workflow's API fields."""
+    if isinstance(workflow, dict):
+        return workflow
+    return {k: v for k, v in workflow.__dict__.items() if not k.startswith('_')}
+
+
+def _spec_model_runtime_image(spec):
+    """Return the model runtime image of an app spec.
+
+    The image is usually at the top level of the spec (`model_runtime_image`)
+    but on some workflows lives nested under `app_type_config`; handle both."""
+    if not isinstance(spec, dict):
+        return None
+    image = spec.get('model_runtime_image')
+    if image is None:
+        app_type_config = spec.get('app_type_config')
+        if isinstance(app_type_config, dict):
+            image = app_type_config.get('model_runtime_image')
+    return image
+
+
+def _spec_thresholds(spec):
+    """Return every threshold-like field of an app spec as a flat dict.
+
+    Threshold fields vary by app type (e.g. `threshold`, `detection_thresholds`)
+    and may live at the top level of the spec or nested under `app_type_config`,
+    so match any key containing 'threshold' in either location."""
+    thresholds = {}
+    if not isinstance(spec, dict):
+        return thresholds
+    containers = [('', spec)]
+    app_type_config = spec.get('app_type_config')
+    if isinstance(app_type_config, dict):
+        containers.append(('app_type_config.', app_type_config))
+    for prefix, container in containers:
+        for key, value in container.items():
+            if 'threshold' in key.lower():
+                thresholds[prefix + key] = value
+    return thresholds
+
+
+def _flatten(value, prefix=''):
+    """Flatten nested dicts into {dotted.path: leaf_value}.
+    Lists and scalars are treated as leaf values (compared whole)."""
+    if isinstance(value, dict) and value:
+        flat = {}
+        for k, v in value.items():
+            path = "%s.%s" % (prefix, k) if prefix else str(k)
+            flat.update(_flatten(v, path))
+        return flat
+    return {prefix: value} if prefix else {}
+
+
+def _value_changes(a, b):
+    """Generic deep compare of two values (typically dicts).
+
+    Returns {dotted.path: change} where change is {'old': ..., 'new': ...};
+    the 'old'/'new' key is omitted when the path is absent on that side."""
+    flat_a = _flatten(a)
+    flat_b = _flatten(b)
+    changes = {}
+    for path in sorted(set(flat_a) | set(flat_b)):
+        old = flat_a.get(path, _MISSING)
+        new = flat_b.get(path, _MISSING)
+        if old == new:
+            continue
+        change = {}
+        if old is not _MISSING:
+            change['old'] = old
+        if new is not _MISSING:
+            change['new'] = new
+        changes[path] = change
+    return changes
+
+
+def summarize_app_spec(spec):
+    """Return a compact one-row summary of a single app spec: application_id,
+    model_runtime_image, and any threshold fields (plus name/type when the
+    workflow actually carries them — they are often null in app_specs)."""
+    if not isinstance(spec, dict):
+        return {'application_id': None, 'model_runtime_image': None}
+    row = {
+        'application_id': spec.get('application_id'),
+        'model_runtime_image': _spec_model_runtime_image(spec),
+    }
+    for key in ('name', 'type', 'app_type'):
+        if spec.get(key) is not None:
+            row[key] = spec[key]
+    thresholds = _spec_thresholds(spec)
+    if thresholds:
+        row['thresholds'] = thresholds
+    return row
+
+
+def workflow_summary(workflow):
+    """Return a compact model-composition summary of a workflow.
+
+    workflow: a CogniacWorkflow / AsyncCogniacWorkflow instance or a plain
+              workflow dict (as returned by GET /1/workflows/{workflow_id}).
+
+    Returns a dict with workflow identity fields plus one summary row per
+    app spec (see summarize_app_spec)."""
+    wf = _as_workflow_dict(workflow)
+    app_specs = wf.get('app_specs') or []
+    return {
+        'workflow_id': wf.get('workflow_id'),
+        'base_id': wf.get('base_id'),
+        'version': wf.get('version'),
+        'name': wf.get('name'),
+        'edgeflow_model': wf.get('edgeflow_model'),
+        'app_count': len(app_specs),
+        'apps': [summarize_app_spec(s) for s in app_specs],
+    }
+
+
+def _workflow_identity(wf):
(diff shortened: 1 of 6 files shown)


wskish (commented):
Reviewed the diff, ran the suite locally, and live-verified against a live tenant.

**Correctness**: the diff keys on `application_id`, surfaces `model_runtime_image` changes prominently (checking both the spec top level and `app_type_config`), and the generic dotted-path deep diff guarantees any field change is caught regardless of location. Positional pairing within an `application_id` group handles duplicate specs sensibly, and missing/None `app_specs` are tolerated. Pure client-side computation with no new endpoints — good call.

**Live verification (read-only)**: diffed two adjacent workflow versions from a real incident on a live tenant. The tool surfaced exactly the one changed app spec, with the `model_runtime_image` retag reported prominently as old/new — zero noise (no spurious adds/removes/top-level entries) across a 100+ app-spec workflow. `workflow summary` on a large version reported 106 app specs, 84 with images; spot-checking the raw workflow confirmed the 22 image-less rows genuinely carry `model_runtime_image: null` (non-model apps), so the summary is accurate.

**One observation from live data (non-blocking)**: real app specs also mirror the image under `app_config.model_runtime_image` (not only `app_type_config`). The generic deep diff caught it (`changed` listed both `model_runtime_image` and `app_config.model_runtime_image`), and the prominent field fired via the top-level copy, so nothing was missed — but consider adding `app_config` to `_spec_model_runtime_image`'s nested lookup so the prominent surfacing also covers a hypothetical spec whose image lives only there.

Also noting `system_software_updates_available` is deliberately excluded from `top_level_changes` as bookkeeping — reasonable, since it flips independently of authored changes; the docstring exclusion list makes the intent clear.

**Tests**: checked out the branch and ran the suite locally — 310 passed, 110 skipped. All fixtures use invented placeholder ids and example registries; no live calls.

cogniac/workflow.py
```diff
@@ -7,6 +7,255 @@
...
+    but on some workflows lives nested under `app_type_config`; handle both."""
+    if not isinstance(spec, dict):
+        return None
+    image = spec.get('model_runtime_image')
+    if image is None:
+        app_type_config = spec.get('app_type_config')
+        if isinstance(app_type_config, dict):
+            image = app_type_config.get('model_runtime_image')
+    return image
+
```
spatel-cog, line 65: **Should-fix: `app_config` is the missing third location, and on at least one large production tenant it is load-bearing, not hypothetical.**

On the tenant I live-verified against (~325 app specs per workflow version), every spec has `name: null` and no `type` at the top level, and `app_type_config` does not exist at the spec level at all. But every spec embeds the full application record under `app_config` -- including `name`, `type`, `detection_thresholds`, `input_threshold`, and a mirrored `model_runtime_image`. Consequences on that tenant:

- `workflow summary` emits rows containing only `application_id` + image -- 0 of ~325 names resolved -- while the name sits unread in `app_config.name`. Issue #182 explicitly asks for the resolved app name so a human can read the composition without joining against `apps get`; that half of the issue stays unmet there.
- `_spec_thresholds` finds nothing (no `*threshold*` keys top-level or in `app_type_config`), so the prominent threshold surfacing never fires; a threshold change appears only in the generic deep diff under an `app_config.*` path.

**Suggested change**: extend `_spec_model_runtime_image`, `_spec_thresholds`, and `summarize_app_spec` (name/type) to fall back to `app_config`. Use fallback, not merge, semantics -- top level wins, then `app_type_config`, then `app_config`. The data shows why merging would be wrong: in my 13-version diff, 9 apps changed the top-level image but 12 changed the `app_config` mirror -- they are not in lockstep (the mirror is a snapshot that churns independently), so merging both into the prominent field would produce spurious old/new pairs, while fallback only activates where the authoritative field is absent. This also matches the earlier review's suggestion here, upgraded with evidence that the gap is real in production data.

spatel-cog (commented):
Review by Claude, on behalf of @spatel-cog. I checked out the branch and live-verified it read-only against a second production tenant (different from the one in the earlier review), picked because its deployment history is deep: several production deployment groups all targeting the same workflow base, currently at 300+ versions with ~325 app specs per version. All identifiers below are scrubbed.

**Live validation against real deployment history:**

- **Adjacent versions (N -> N+1)**: the diff reported exactly 4 changed apps, each a single `model_runtime_image` retag, plus the authored name/description change -- zero noise across ~325 specs. The workflow's own authored description for version N+1 says the AI models were updated, so the tool's output corroborates the authored history. This is precisely the incident shape issue #182 describes.
- **A 13-version jump**: 4 apps added, 38 changed -- consistent with the intervening authored descriptions. Compute cost is negligible (~0.3s CPU; wall time is the two API fetches).
- **Fit with the deployment model**: deployment groups store only the current `target_workflow_id`, so reconstructing "what changed when" is exactly adjacent-version diffing -- which this enables directly from the `base:version` ids that `deployments list` / `workflows list` already expose. Version-0 workflows (`base:0`) exist in the wild and are handled fine.

**One requested change before merge**: on this tenant, the spec top level and `app_type_config` are not where the interesting fields live -- everything meaningful (name, type, thresholds, a mirrored image) is nested under `app_config`, which the prominent-field helpers and the summary never consult. This upgrades the earlier review's non-blocking `app_config` suggestion to a should-fix, and it extends beyond the image field. Details and evidence on the inline comment in `workflow.py`; a companion inline in the tests suggests the regression fixture.

**Non-blocking observation**: on multi-version jumps the generic `changed` dict picks up `app_config` snapshot churn (`app_config.app_type_config.build_id` on 20 apps in my 13-version window, plus `modified_at`, `hpo_credit`, `model_id`, ...). That is by design ("catch everything"), and on adjacent versions it was silent. I lean toward keeping the behavior and just noting in the `workflow_diff` docstring that `changed` includes app-record snapshot bookkeeping -- exhaustiveness is the feature; a spec-level exclusion list (parallel to `_DIFF_EXCLUDED_TOP_LEVEL`) is the alternative if it proves noisy in practice.

Everything else checks out: structure (reusable module functions, prominence layered over a generic deep diff so nothing is silently missed), `_as_workflow_dict` safety (`__init__` stores only `_`-prefixed non-API attributes -- verified), the documented list-as-leaf and positional-pairing tradeoffs (no duplicate `application_id` groups in the production workflow I checked), offline test coverage, and public-repo hygiene (fixtures and PR body are clean, and this review is scrubbed accordingly).

LGTM once the `app_config` fallback lands.

tests/test_workflow_diff.py
```diff
@@ -0,0 +1,292 @@
...
+
+import cogniac
+from cogniac.workflow import CogniacWorkflow, workflow_diff, workflow_summary
+
+
+# ---------------------------------------------------------------------------
+# fixture workflows (invented placeholder ids — no real tenant data)
+# ---------------------------------------------------------------------------
+
+def make_workflow_a():
```
spatel-cog, line 19: The fixtures model the two documented spec shapes (fields at the top level, and nested under `app_type_config`), but not the third shape observed in production: top-level `name`/`type` null, no `app_type_config` at the spec level, and everything meaningful (name, type, `detection_thresholds`, `input_threshold`, mirrored `model_runtime_image`) nested under `app_config`. If the `app_config` fallback from the `workflow.py` comment lands, please add one fixture app in that shape -- it becomes the regression guard proving the summary resolves names (and the prominent threshold/image fields fire) without an API join, on the spec shape where it currently cannot.

cogniac/workflow.py
wskish: Landed in 8d2b8c4. `_spec_model_runtime_image`, `_spec_thresholds`, and `summarize_app_spec` (name/type) now fall back to the embedded `app_config` record with the precedence you specified — top level → `app_type_config` → `app_config`, fallback not merge, so the independently-churning mirror is only consulted where the authoritative locations lack the field (there's a regression test asserting mirror-only churn does not fire the prominent field). Also added the docstring note that the generic `changed` dict intentionally includes app-record snapshot bookkeeping on multi-version jumps, and `workflow_diff`/`workflow_summary` now report `app_specs_missing: true` when a record omits `app_specs` entirely (versions-list summary shape) instead of silently showing zero apps.

tests/test_workflow_diff.py
wskish: Added in 8d2b8c4: a fixture app in exactly this third shape (top-level name/type null, no spec-level `app_type_config`, name/type/image/`detection_thresholds`/`input_threshold` under `app_config`), with tests proving the summary resolves the name and the prominent image/threshold surfacing fires without an API join — plus a stale-mirror test guarding the fallback-not-merge semantics.

PR #189, opened 2026-07-15
Add client-derived EdgeFlow fleet health: `edgeflows health` CLI + SDK health()/get_all_health()

wskish:
Addresses the client-derivable part of #184 (`edgeflows list/get` return null `last_seen`/`connection_status`, making fleet health invisible from the CLI). Fully populating connectivity on the gateway record itself needs backend work, so #184 stays open — but the per-device status endpoint works today, and each status record carries `cc_timestamp` (cloud-receipt clock), so an effective health summary can be derived client-side.

## What's added

**CLI**

```
cogniac edgeflows health [--stale-minutes N]     # default 15
```

Emits one JSON record per EdgeFlow in the tenant:

```json
{"gateway_id": ..., "name": ..., "deployment_group_id": ..., "last_seen": <epoch|null>, "online": <bool>, "current_workflow_id": ...}
```

- `last_seen` is the `cc_timestamp` of the device's most recent status record (fetched with `reverse=True&limit=1`, one bounded GET per device, fetched concurrently on up to 8 threads); falls back to `timestamp`/`gw_timestamp` on records that lack it.
- `online` = `last_seen` within `--stale-minutes` of now.
- A device with zero status records reports `last_seen: null, online: false`.
- `--format table` supported via new `edgeflow_health` columns; `gateway health` etc. work via the existing synonym groups.

**SDK** (sync + async, per the repo's symmetry rule)

- `CogniacEdgeFlow.get_all_health(connection, stale_seconds=900, max_workers=8)` — tenant-level summary (what the CLI uses).
- `CogniacEdgeFlow.health(stale_seconds=900)` — one device.
- `AsyncCogniacEdgeFlow.get_all_health(...)` / `.health(...)` — async counterparts (`asyncio.gather` with a semaphore cap), sharing the same derivation helper.

## Honest caveat (in docstrings and --help)

This is a **client-side derivation from status records on the cloud-receipt clock, not a true device heartbeat**: a device uploading a backlog of status records can briefly look "online" while the backlog drains. The real fix for `list`/`get` nulls remains backend-side.

## Tests

Mocked connections only (no live API calls): online / stale / no-records cases, `cc_timestamp` preferred over `gw_timestamp` (with `gw_timestamp` fallback), empty tenant, per-device request boundedness (`limit=1&reverse=True`), the `--stale-minutes` flag, instance-method field passthrough, and async parity. The smoke-test pairing lists gain `health` / `get_all_health`. Full suite: **300 passed, 110 skipped** (skips are the live-credential tests).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -182,7 +182,9 @@ cogniac edgeflows health        # client-derived fleet health summary: per devic
                                 # recent status record; --stale-minutes (default 15) sets the online
                                 # threshold. Not a true heartbeat: the backend does not populate
                                 # connectivity on the gateway record, and a device uploading backlogged
-                                # status can briefly look online.
+                                # status can briefly look online. online is tri-state: true/false, or
+                                # null when that device's status fetch failed (its record then carries
+                                # an `error` key; one bad device never aborts the sweep).
 cogniac edgeflows metrics names              # list available metric names
 cogniac edgeflows metrics list --metric-name <name>   # time-series for a metric; --edgeflow-id scopes to one device,
                                 #   --start/--end (epoch or ISO 8601, supplied together) bound the window
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -176,6 +176,13 @@ cogniac edgeflows status <id>   # status events: --subsystem, --start, --end, --
                                 # --start/--end filter on the cloud-receipt timestamp (cc_timestamp),
                                 # not device time (clocks can diverge on a skewed/backfilling EdgeFlow);
                                 # --sort is not exposed (backend defaults to cloudcore_timestamp)
+cogniac edgeflows health        # client-derived fleet health summary: per device {gateway_id, name,
+                                # deployment_group_id, last_seen, online, current_workflow_id}.
+                                # last_seen = cc_timestamp (cloud-receipt clock) of the device's most
+                                # recent status record; --stale-minutes (default 15) sets the online
+                                # threshold. Not a true heartbeat: the backend does not populate
+                                # connectivity on the gateway record, and a device uploading backlogged
+                                # status can briefly look online.
 cogniac edgeflows metrics names              # list available metric names
 cogniac edgeflows metrics list --metric-name <name>   # time-series for a metric; --edgeflow-id scopes to one device,
                                 #   --start/--end (epoch or ISO 8601, supplied together) bound the window
--- a/README.md
+++ b/README.md
@@ -138,6 +138,7 @@ cogniac media get --media-id <id>                  # media metadata
 cogniac media download --media-id <id> -o out.jpg  # download media to file
 cogniac edgeflow list                              # list edge devices
 cogniac edgeflow status --edgeflow-id <id> --list-subsystems   # discover which subsystems a device reports
+cogniac edgeflow health                            # fleet health derived from each device's latest status record
 cogniac application create --body @app.json        # --body takes inline JSON, @FILE, or - (stdin)
 ```
 
--- a/cogniac/async_edgeflow.py
+++ b/cogniac/async_edgeflow.py
@@ -4,9 +4,23 @@ Async CogniacEdgeFlow Object Client
 Copyright (C) 2016 Cogniac Corporation
 """
 
+import asyncio
 import os
 from time import time
 from .common import retry, stop_after_attempt, wait_exponential, retry_if_exception, server_error
+from .edgeflow import DEFAULT_HEALTH_STALE_SECONDS, _health_record
+
+
+@retry(stop=stop_after_attempt(8), wait=wait_exponential(multiplier=0.5), retry=retry_if_exception(server_error))
+async def _latest_status_record(connection, gateway_id):
+    """
+    Return the most recent status record for a gateway, or None if the device
+    has never reported status. One bounded GET (reverse-sorted, limit 1)
+    against GET /1/gateways/{gateway_id}/status — no pagination walk.
+    """
+    resp = await connection._get("/1/gateways/%s/status?reverse=True&limit=1" % gateway_id)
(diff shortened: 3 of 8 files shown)


wskish (commented):
Reviewed the diff, ran the suite locally, and live-verified against a live tenant fleet.

**Correctness**
- Devices with zero status records are handled per spec (`last_seen: null, online: false`) and missing gateway fields surface as null via `.get()`/`getattr(..., None)` rather than crashing.
- The `cc_timestamp` (cloud-receipt clock) preference with `timestamp`/`gw_timestamp` fallback is the right call, and the docstrings/help are refreshingly honest that this is a derivation, not a heartbeat.
- One bounded `reverse=True&limit=1` GET per device matches how the existing `status()` builds its query string, and the per-request `@retry(server_error)` covers 429 backoff during the fan-out.
- `connection.tenant_id` is populated on both sync and async connections, and `httpx.Client` is thread-safe, so the ThreadPoolExecutor fan-out over a shared connection is sound.

**Convention fit**: CLI verb registration, `--format table` columns, sync/async symmetry with a shared derivation helper, and smoke-test pairing list updates all match repo patterns.

**Tests**: checked out the branch and ran the suite locally — 300 passed, 110 skipped. Mock-only, placeholder ids throughout, and the tests assert the request boundedness (limit=1) which guards against a future accidental full-history walk.

**Live verification (read-only)**: ran `edgeflows health` against a live tenant fleet of 66 devices — completed in under 5 seconds with zero errors. Every gateway independently confirmed active in the last hour reported `online: true` with `last_seen` under a minute old; the handful of quiet devices reported `online: false` with plausibly stale `last_seen` values (hours to months). No inconsistent records (nothing online with a stale/absent last_seen), and every record had the full documented shape.

Minor, non-blocking: `_health_record`'s `or`-chained timestamp fallback would treat a literal `cc_timestamp: 0` as missing — harmless in practice (epoch-zero would be offline either way), just noting it.

spatel-cog (commented):
LGTM overall — the design premise holds up and the implementation is convention-clean. Two findings below worth addressing before merge.

## Verification done

- Re-ran the suite on the branch with credentials scrubbed: **300 passed, 110 skipped**, matching the PR description; all 10 new health tests pass.
- Confirmed the backend premise: the status store's range key is the cloud-receipt timestamp, so `reverse=True&limit=1` genuinely returns the newest record by cloud receipt, and the route's default sort matches.
- Empirically spot-checked production devices: healthy devices post status records roughly every minute (the 1-minute GPU timer plus other subsystems), so the 15-minute default staleness threshold is conservative and sane. Devices that are gone show days-to-years of silence. The liveness derivation is sound.
- Conventions check out: sync/async symmetry, `server_error` retry layered on `_get` per the repo rule, `reverse=True` string form matching the existing `status` method, list return matching `get_all`, error envelope, smoke-pairing lists updated, placeholder-only test IDs (right call for a public repo).

## Finding 1: one bad device aborts the whole fleet sweep

`get_all_health` propagates the first per-device exception and loses the entire summary:

- sync (`cogniac/edgeflow.py`): `pool.map(one, gateways)` re-raises the first exception.
- async (`cogniac/async_edgeflow.py`): `asyncio.gather(...)` without `return_exceptions` does the same.

A gateway deleted between the tenant list call and its status GET returns 404 → `ClientError` → not retried → the whole sweep raises. Same for a single device whose status GET exhausts the 5xx retries. This bites hardest on exactly the large fleets this command exists for.

Suggest wrapping the per-device fetch in try/except and emitting a degraded record, e.g. `{gateway_id, name, ..., last_seen: null, online: null, error: "<msg>"}` (or `online: false` — but `null` distinguishes "couldn't determine" from "determined offline"). Plus a test pinning that behavior — it's undefined today.

## Finding 2: `connection.tenant_id` vs `connection.tenant.tenant_id`

`get_all_health` uses `connection.tenant_id` where the sibling `get_all` uses `connection.tenant.tenant_id`. On a tenant-less connection (API key with no tenant selected, or multi-tenant user without `COG_TENANT`), the `tenant` property raises the SDK's helpful "must specify tenant" error listing available tenants, while the raw attribute is `None` and produces an opaque `/1/tenants/None/gateways` 4xx. Use the property for parity and the better failure mode. (The mocked test connections set `tenant_id` directly, which is why the tests can't see the difference.)

---
🤖 Review generated with [Claude Code](https://claude.com/claude-code)

wskish:
Both findings addressed in e6901b8:

**Finding 1 (one bad device aborts the sweep)**: the per-device status fetch is now wrapped in try/except on both paths (sync `ThreadPoolExecutor` map and async `gather` coroutine). A failing device emits a degraded record `{gateway_id, name, deployment_group_id, last_seen: null, online: null, error: "<msg>"}` while every other device still reports. `online` is now tri-state — `null` means "could not determine" (fetch failed), distinct from `false` "determined offline" — documented in the SDK docstrings, CLI `--help`, and CLAUDE.md; the `edgeflow_health` table format gains an `error` column. New tests on both sync and async paths pin the behavior: one device raising a 404 `ClientError` mid-sweep degrades only its own record, with shape and gateway-field passthrough asserted.

**Finding 2 (`tenant_id` vs `tenant.tenant_id`)**: sync `get_all_health` now uses `connection.tenant.tenant_id`, matching sibling `get_all`, so a tenant-less connection gets the SDK's helpful "must specify tenant" error instead of an opaque `/1/tenants/None/gateways` 4xx. The sync test mock now exposes the tenant id *only* via the `tenant` property path, so a regression back to the raw attribute fails the tests. One note: the async version keeps `connection.tenant_id` — that *is* parity with its sibling `AsyncCogniacEdgeFlow.get_all`, since `AsyncCogniacConnection` has no `tenant` property; the async mock documents this.

Full suite: 302 passed, 110 skipped (mock-only, placeholder ids).

PR #192, opened 2026-08-23
Log the resolved url_prefix instead of the raw parameter

bjw-0:
## What
One-line fix in `CogniacConnection.__init__`: the "Connecting to Cogniac system at ..." log line prints the raw `url_prefix` **parameter** instead of the resolved `self.url_prefix`.

## Why
`__init__` resolves the URL into `self.url_prefix` (explicit argument → `COG_URL_PREFIX` env → default), but logs the parameter. Every caller that relies on the env var or the default — e.g. `CogniacConnection(tenant_id=...)` — logs `Connecting to Cogniac system at None` on **every** connection, successful ones included. During incident triage this reads as a broken FQDN/configuration when the connection is actually fine; it cost real time today distinguishing this cosmetic line from an unrelated failure in the same log.

## Scope
Log line only. No behavior change — the resolved value was already used for all requests.

## Verification
Grepped the file for other uses of the raw parameter after resolution: this log line is the only one. The same defect pattern exists in services pinned to older SDK versions; they pick up the fix on their next pin bump.


bjw-0:
@spatel-cog small logging fix (log the resolved url_prefix, not the raw parameter), quiet 18 days. Ready when you are.

Agent: claude-fable-5-1 | claude-code-2.1.267

spatel-cog (approved):

